PR #95, opened 2022-11-07
documentation: update readme

ncitron:
- mention eth_getProof requirement for eth1 rpc (thanks @lightclient for pointing this out)
- italicize disclaimer

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,9 @@ To run Helios, run the below command, replacing `$ETH_RPC_URL` with an RPC provi
 ```
 helios --execution-rpc $ETH_RPC_URL
 ```
+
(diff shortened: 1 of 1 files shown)


ryanschneider:
@ncitron not sure where you got the information that Infura doesn't support `eth_getProof` but it does.

lightclient:
@ryanschneider it was from me. At the time, [the docs weren't clear](https://web.archive.org/web/20220523110744/https://docs.infura.io/infura/networks/ethereum/json-rpc-methods) that infura supported `eth_getProof`. Sorry for the misinformation!

ncitron:
@ryanschneider it does however seem that infura does not support `eth_createAccessList` though which helios requires. I will however update the readme to reflect this. Thanks for pointing this out.

PR #97, opened 2022-11-08
refactor: add count parameter to get_update and use 128 as the limit

dadepo:
This PR updates the `get_updates` making it accept count. It also limit the count used in request to 128 as specified in the specification https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#configuration instead of hard coding to 1000


common/src/constants.rs
```diff
@@ -0,0 +1,4 @@
+// Consensus constants
+
+// https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#configuration
+pub const MAX_REQUEST_LIGHT_CLIENT_UPDATES: u8 = 128;
```
ncitron, line 4: Do you anticipate this constant needing to be exposed to any other package other than consensus? If not, can we move the constants module into the consensus package?

Other than that looks good

ncitron:
Thanks for the contribution!

PR #100, opened 2022-11-09
fix: explicit check for invalid divergent paths

### LINKED ISSUE #99: bug: when verifying proofs, shared_prefix_length doesn't exit when a non-matching nibble is found
Hi! I noticed the shared_prefix_length function in the proofs module compares the paths from the proof and the path we're verifying, but it doesn't exit early if a nibble that doesn't match is found: 

https://github.com/a16z/helios/blob/master/execution/src/proof.rs#L136-L138

For instance, when comparing `0xabcd` with `0xabfd` it will return a shared prefix length of 3 instead of 2.

This means a proof could actually be showing a divergent path, but because some nibbles match at the end of the path, the verifier will advance to later nibbles and potentially validate an invalid proof (though this is probably highly impractical to exploit in a real-world account or storage proof).

Moreover, as we're walking along the proof, if such a divergent path were encountered, the verifier should reject the proof if this is not the last node.

(I noticed this while comparing the verifier implementation to https://github.com/lidofinance/curve-merkle-oracle/blob/main/contracts/MerklePatriciaProofVerifier.sol#L74-L95 as it has a similar structure)

I have a fix for this ready in https://github.com/pcarranzav/helios/tree/fix-shared-prefix-length and will PR right after posting this (my laptop just seems to be taking ages to install the dependencies so that I can check the tests run properly...)

pcarranzav:
Fixes #99 

i.e. it now properly rejects proofs where the paths diverge in an extension node.

I also refactored shared_prefix_length a little bit to avoid looping through the node path twice. I added a regression test for the prefix length calculation; testing that the proofs are rejected if the divergent path is found before the end is a bit more complex as I'd have to actually construct a proof like that.

(I still haven't managed to install the deps so that I can actually run the tests, so please bear with me if something doesn't quite build or pass yet - I'll adjust as soon as possible, but I wanted to prioritize PRing the fix)

### DIFF
--- a/execution/src/proof.rs
+++ b/execution/src/proof.rs
@@ -53,7 +53,7 @@ pub fn verify_proof(proof: &Vec<Bytes>, root: &Vec<u8>, path: &Vec<u8>, value: &
             } else {
                 let node_path = &node_list[0];
                 let prefix_length = shared_prefix_length(path, path_offset, node_path);
-                if prefix_len < node_path.len() * 2 - skip_length(node_path) {
+                if prefix_length < node_path.len() * 2 - skip_length(node_path) {
                     // The proof shows a divergent path, but we're not
                     // at the end of the proof, so something's wrong.
                     return false;
--- a/execution/src/proof.rs
+++ b/execution/src/proof.rs
@@ -177,7 +177,7 @@ mod tests {
     use crate::proof::shared_prefix_length;
 
     #[tokio::test]
-    fn test_shared_prefix_length() {
+    async fn test_shared_prefix_length() {
         // We compare the path starting from the 6th nibble i.e. the 6 in 0x6f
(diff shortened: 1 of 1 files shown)


ncitron:
Thanks for the contribution! 

I think technically this cannot be exploited since any attempt to bypass the rejection will still always fail in the length check we perform at the end of the proof. It does however increase the correctness of our verify function, and adds extra guardrails, so will merge. I also appreciate the refactor of `shared_prefix_length`.

pcarranzav:
Thanks for merging!
I agree it'd probably have been impossible to exploit, because of what you mention, but also because you need paths that are very similar, and similar keccak256 hashes should be hard to find in the first place. Glad you found it useful nonetheless.

Issue #98, opened 2022-11-08
check checkpoint age when starting up


ncitron:
When Helios starts, it needs to fetch a recent beacon blockhash to use as the weak subjectivity checkpoint. If this blockhash is too old (under worst case conditions too old is ~14 days), it is possible for an attacker to trick Helios into following the wrong chain. While this attack is hard to pull off (requires millions in capital to fill the staking deposit and withdrawal queues), we should still check the checkpoint age, and if it is too old, throw and error and tell the user how to fetch a good blockhash.

To do this, use the bootstrap fetched [here](https://github.com/a16z/helios/blob/master/consensus/src/consensus.rs#L150) and check `bootstap.header.slot`'s age using the `expected_current_slot` and `slot_timestamp` methods in `Consensus`.

If it is older that 14 days, throw an error (consensus errors can be found [here](https://github.com/a16z/helios/blob/master/consensus/src/errors.rs))

dadepo:
> If this blockhash is too old (under worst case conditions too old is ~14 days)

@ncitron Is the age for the worst case condition static and always be ~14 days? 

From my quick skim through this [post](https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2#Weak-Subjectivity-Period) it looks it varies and depends on validator set, entry and exit rates...things that might be difficult for helios to attain without having access to the consensus state.

ncitron:
Yeah it does change, but looking at the bottom chart, it seems like 14 days is a good metric since it increases with the validator set size. Even at 250k its 2 weeks, and Ethereum mainnet has 500k validators. It does bring up a good point that this should be configurable. @0xModene can we add it as a network level config options in `config/src/networks.rs` instead of hardcoding it to 14 days?

0xModene:
Sure thing! I'll get to work on this

ncitron:
closed in #105

PR #105, opened 2022-11-11
feat: verify checkpoint has valid age

0xModene:
Addressing issue #98 

Verifies that the blockhash isn't too old (14 days or older according to the issue) by checking timestamps against the current slot. If it's too old, a timestamp error is thrown

note: also removed `#![feature(map_first_last)]` in `lib.rs` as there were some compiler warnings thrown that this has been implemented in stable releases and is no longer required

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -166,6 +166,7 @@ impl ClientBuilder {
             data_dir,
             chain: base_config.chain,
             forks: base_config.forks,
+            checkpoint_duration: base_config.checkpoint_duration,
         };
 
         Client::new(config)
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -166,7 +166,7 @@ impl ClientBuilder {
             data_dir,
             chain: base_config.chain,
             forks: base_config.forks,
-            checkpoint_duration: base_config.checkpoint_duration,
+            max_checkpoint_age: base_config.max_checkpoint_age,
         };
 
         Client::new(config)
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -507,7 +507,7 @@ impl<R: ConsensusRpc> ConsensusClient<R> {
 
         let slot_age = current_slot_timestamp - blockhash_slot_timestamp;
 
-        slot_age < self.config.checkpoint_duration
+        slot_age < self.config.max_checkpoint_age
     }
 }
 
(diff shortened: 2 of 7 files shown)


consensus/src/consensus.rs
```diff
@@ -153,6 +154,11 @@ impl<R: ConsensusRpc> ConsensusClient<R> {
             .await
             .map_err(|_| eyre!("could not fetch bootstrap"))?;
 
+        let is_valid = self.is_valid_blockhash(bootstrap.header.slot);
+        if !is_valid {
+            return Err(ConsensusError::InvalidTimestamp.into());
```
ncitron, line 159: Can we add a new error called `CheckpointToOld` instead so we can provide better error messages?

consensus/src/consensus.rs
```diff
@@ -493,6 +499,17 @@ impl<R: ConsensusRpc> ConsensusClient<R> {
 
         Duration::seconds(next_update as i64)
     }
+
+    // Determines blockhash_slot age and returns true if it is less than 14 days old
+    fn is_valid_blockhash(&self, blockhash_slot: u64) -> bool {
```
ncitron, line 504: can we rename this `is_valid_checkpoint`?

consensus/src/consensus.rs
```diff
@@ -6,6 +6,7 @@ use blst::min_pk::PublicKey;
 use chrono::Duration;
 use eyre::eyre;
 use eyre::Result;
+use log::warn;
```
ncitron, line 9: Is this import being used?

ncitron (commented):
Looks pretty good! Just a few minor nitpicks. 

Also make sure that it is linted using `cargo fmt` and tested with `cargo test --all`

ncitron:
Also see: https://github.com/a16z/helios/issues/98#issuecomment-1311965962

consensus/src/consensus.rs
0xModene: Nope, used it in testing to get timestamp vals and forgot to remove.

0xModene:
@ncitron mind checking again?

client/src/client.rs
```diff
@@ -166,6 +166,7 @@ impl ClientBuilder {
             data_dir,
             chain: base_config.chain,
             forks: base_config.forks,
+            checkpoint_duration: base_config.checkpoint_duration,
```
ncitron, line 169: can we rename this to something more descriptive, like `max_checkpoint_age`. Other than that LGTM.

0xModene: done

Issue #23, opened 2022-09-12
implement logging rpc methods


ncitron:
We currently don't have any of the rpc methods related to fetching logs implemented. There seems to be a lot of them, and I'm not sure which ones dapps use the most (it seems dapps are using logs less and less nowadays). I think the best plan is to use a bunch of dapps and keep track of which of the log methods they are using and implement them.

AlbionaHoti:
Hi @ncitron, I'm new to rust in general, but I just saw the project and good first issue tag, and I'd like to contribute in any way. I'll start researching what rpc methods other dapps use for fetching logs.

simon-saliba:
Hey, I opened a draft PR. I'm not sure if we need some kind of "proof" for logs, as it is done for accounts for example.

wighawag:
+1 for eth_getLogs and chiming it to say that logs are a very important mechanism for fully decentralized apps

@ncitron 
>  (it seems dapps are using logs less and less nowadays)

This only because more and more web3 app relies on third party servers but truly decentralized app will make use of eth_getLogs themselves

Currently working on in-browser generic indexer to faciliate this, see : https://github.com/conquest-eth/ethereum-indexer

PR #108, opened 2022-11-12
feat: add `get_logs` RPC method

simon-saliba:
Fixes issue: https://github.com/a16z/helios/issues/23

This PR is still in draft.

I was wondering if we need any kind of "proof" for logs as we do for account details ?

### DIFF
--- a/execution/src/execution.rs
+++ b/execution/src/execution.rs
@@ -21,6 +21,10 @@ use super::proof::{encode_account, verify_proof};
 use super::rpc::ExecutionRpc;
 use super::types::{Account, ExecutionBlock};
 
+// We currently limit the max number of logs to fetch,
+// to avoid blocking the client for too long.
+const MAX_SUPPORTED_LOGS_NUMBER: usize = 5;
+
 #[derive(Clone)]
 pub struct ExecutionClient<R: ExecutionRpc> {
     pub rpc: R,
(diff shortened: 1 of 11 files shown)


ncitron:
This is a very good start. We definitely do need a mechanism to prove that the log result is correct though. I think the way to do this would be to use receipt proofs (which we already do to implement all the getReceipt methods). We can go through the transactions that produced the returned logs, then fetch and prove the receipt for them (can just use the receipt proving infra we already have). We can then check that receipt includes the log we exepcted.

simon-saliba:
Thanks for the explanation ! 

I'm trying to understand the proofing mechanism for transaction receipts.
We fetch the receipt of a certain tx from the untrusted rpc source. On the other hand, we fetch all of the receipts for all the transactions inside the same block and locally build a receipts root. 
Then we compare the receipts root we calculated locally to the receipts root in the execution payload, if I understand correctly. 

My questions: 

1- What makes us trust the `payloads` field inside the `Node` struct ?
2- If the list of receipts contain our "to-be-proved" receipt, and this list of receipts reconstructs the same receipts root in the execution paylaod, is it enough to conclude that the fetched receipt is correct?

ncitron:
Yeah your understanding is correct.

1. The payloads are all proved out by the `consensus` package. We can assume that any payload used inside of `execution` is valid.
2. Yes

simon-saliba:
I added a first version of the approach we discussed, WDYT ?

For the test I used the same log inside the `transaction.json`, but copied to a separate file. Not sure if it's the best approach versus reading `transaction.json` and returning `tx.logs[0]` inside `MockRPC`.

simon-saliba:
I'm getting this error 
```
the trait `Deserialize<'_>` is not implemented for `&ethers::types::Filter`
```

However, when you look at the `Filter` type in `ethers-core`, it seems to me that the trait is implemented: [Filter](https://docs.rs/ethers-core/latest/ethers_core/types/struct.Filter.html#impl-Deserialize%3C%27de%3E-for-Filter)

ncitron:
It may be because it is implemented only for Filter and not &Filter. Try just hating Filter for the jsonrpc server handling.

execution/src/execution.rs
```diff
@@ -262,6 +262,45 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
 
         Ok(Some(tx))
     }
+
+    pub async fn get_logs(
+        &self,
+        filter: &Filter,
+        payloads: &BTreeMap<u64, ExecutionPayload>,
+    ) -> Result<Vec<Log>> {
+        let logs = self.rpc.get_logs(filter).await?;
```
ncitron, line 271: So right now proving a single receipt's existence is very slow. We are working on a potential change to the Ethereum RPC to make it much more efficient, but for now we just have to live with it. A concern I have is that users will make a query for tons of logs, and then Helios will be stuck processing this request for many minutes.

I think a good solution for now would be to limit the number of logs that can be returned to say ~5 logs. Alchemy actually does this, but with a max of 10k logs instead.

The question is whether we should simply truncate the logs or throw an error. I'm leaning towards throwing an error since I don't want Helios to ever return incomplete data.

execution/src/execution.rs
```diff
@@ -262,6 +262,45 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
...
+
+        for (_pos, log) in logs.iter().enumerate() {
+            // For every log
+            // Get the hash of the tx that generated it
+            let tx_hash = log.transaction_hash.unwrap();
```
ncitron, line 276: Lets avoid doing a naked `unwrap` to prevent it from panicing. Take a look at the `ok_or` function method which converts options to results.

ncitron (commented):
Mostly looking good

execution/src/execution.rs
simon-saliba: > The question is whether we should simply truncate the logs or throw an error. I'm leaning towards throwing an error since I don't want Helios to ever return incomplete data.

totally agree

ncitron:
LGTM! Thanks for the contribution!

PR #111, opened 2022-11-23
Add ZMOK RPC provider

tomasbanik:
ZMOK now supports the eth_getProof method. As we really like the Helios project, hope ZMOK users will start using it too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To run Helios, run the below command, replacing `$ETH_RPC_URL` with an RPC provi
 helios --execution-rpc $ETH_RPC_URL
 ```
 
-`$ETH_RPC_URL` must be an Ethereum provider that supports the `eth_getProof` endpoint. Infura does not currently support this. We recommend using Alchemy or [ZMOK](https://zmok.io/).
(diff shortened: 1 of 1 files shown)


makemake-kbo:
subtle advertising 🙄

tomasbanik:
> very subtle 🙄

:) heh ok, well, how to edit it to make it normal/acceptable?

tomasbanik:
I deleted the link to ZMOK too. @makemake-kbo please consider if it still sounds subtle or not. Thanks.

ncitron:
For now, I think we will refrain from adding additional RPC providers as examples in the readme. In the future, I may consider maintaining a list of providers that support the requires methods, and will be sure to include ZMOK if we do.

tomasbanik:
> For now, I think we will refrain from adding additional RPC providers as examples in the readme. In the future, I may consider maintaining a list of providers that support the requires methods, and will be sure to include ZMOK if we do.

sure thing. I just saw Infura so thought you already do additional RPC providers. But if you mention exclusively a16z portfolio companies, we understand that. And as you said, maybe one day you will add other RPC providers (or we will become a part of the a16z portfolio ;), who knows what comes first)

PR #115, opened 2022-11-29
feat: Lints, Clippy, and Cleaning

abigger87:
## Overview

Fixes all linting errors raised by nightly with
```bash
cargo check --all && cargo +nightly fmt -- --check && cargo +nightly clippy --all --all-features -- -D warnings && cargo test --all --all-features
```

Additionally removes unused cargo packages in subcrates.

This is preparatory work for further contributions @ncitron  😄

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -53,7 +53,7 @@ impl Config {
             Err(err) => {
                 match err.kind {
                     figment::error::Kind::MissingField(field) => {
-                        let field = field.replace("_", "-");
+                        let field = field.replace('_', "-");
 
                         println!(
                             "\x1b[91merror\x1b[0m: missing configuration field: {}",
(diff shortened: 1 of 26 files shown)


config/src/lib.rs
```diff
@@ -53,7 +53,7 @@ impl Config {
             Err(err) => {
                 match err.kind {
                     figment::error::Kind::MissingField(field) => {
-                        let field = field.replace("_", "-");
+                        let field = field.replace('_', "-");
```
ncitron, line 56: Should both parameters using single quotes here?

ncitron (commented):
Thanks for the work refactoring here! I'm curious, is it generally considered a good practice to prefer slices over vectors as function inputs since most collections can be easily coerced into a slice?

Also, can we update CI to use the new linting rules?

config/src/lib.rs
abigger87: Nope, since the second argument is expected to be a `&str` substring, whereas the first is generic. See https://doc.rust-lang.org/std/string/struct.String.html#method.replace

Screenshot 2022-11-29 at 12 25 14 PM

ncitron (approved):
LGTM

HAOYUatHZ:
> Also, can we update CI to use the new linting rules?

https://github.com/a16z/helios/pull/122

PR #118, opened 2022-11-30
chore: export CallOpts type

### LINKED ISSUE #117: `CallOpts` type is not exported, hence we cannot use `call` function of Helios client
## Description

Currently `execution::types::CallOpts` is not exported, and it is needed to be able to use `call` function of Helios client.

Will submit a PR.

abdelhamidbakhta:
## Description

Currently execution::types::CallOpts is not exported, and it is needed to be able to use call function of Helios client.

## Changes summary

- Updated Helios `lib.rs` to export `execution::types::CallOpts`

Closes #117

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,7 @@ pub mod config {
 
 pub mod types {
     pub use common::types::BlockTag;
+    pub use execution::types::CallOpts;
 }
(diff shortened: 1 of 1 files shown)


ncitron (approved):
LGTM!

PR #120, opened 2022-11-30
feat: Checkpoint Fallbacks [RFC]

abigger87:
## Overview

Introduces lazy checkpoint fallback services.

### CLI Flags

The following two cli flags are introduced in this pr.

- `--fallback` or `-f` sets the checkpoint fallback url (a string). This is only used if the checkpoint provided by the `--checkpoint` flag is too outdated for Helios to use to sync.
If none is provided and the `--load-external-fallback` flag is not set, Helios will error.
For example, you can specify the fallback like so: `helios --fallback "https://sync-mainnet.beaconcha.in"` (or using shorthand like so: `helios -f "https://sync-mainnet.beaconcha.in"`)
- `--load-external-fallback` or `-l` enables weak subjectivity checkpoint fallback (no value needed).
For example, say you set a checkpoint value that is too outdated and Helios cannot sync to it.
If this flag is set, Helios will query all network apis in the community-maintained list
at [ethpandaops/checkpoint-synz-health-checks](https://github.com/ethpandaops/checkpoint-sync-health-checks/blob/master/_data/endpoints.yaml) for their latest slots.
The list of slots is filtered for healthy apis and the most frequent checkpoint occuring in the latest epoch will be returned.
Note: this is a community-maintained list and thus no security guarantees are provided. Use this is a last resort if your checkpoint passed into `--checkpoint` fails.
This is not recommened as malicious checkpoints can be returned from the listed apis, even if they are considered _healthy_.
This can be run like so: `helios --load-external-fallback` (or `helios -l` with the shorthand).

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -7,7 +7,7 @@ use ethers::types::{Filter, Log, Transaction, TransactionReceipt, H256};
 use eyre::{eyre, Result};
 
 use common::types::BlockTag;
-use config::Config;
+use config::{CheckpointFallback, Config};
 use consensus::{types::Header, ConsensusClient};
 use execution::types::{CallOpts, ExecutionBlock};
 use log::{info, warn};
@@ -210,10 +210,7 @@ impl<DB: Database> Client<DB> {
             let fallback = if let Some(fallback) = &self.fallback {
                 info!("attempting to load checkpoint from fallback {}", fallback);
 
-                let checkpoint =
-                    config::checkpoints::CheckpointFallbackList::fetch_checkpoint_from_api(
-                        fallback,
-                    )
+                let checkpoint = CheckpointFallback::fetch_checkpoint_from_api(fallback)
                     .await
                     .map_err(|_| {
                         eyre::eyre!("Failed to fetch checkpoint from fallback: {}", fallback)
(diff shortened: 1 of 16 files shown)


client/src/client.rs
```diff
@@ -166,7 +202,82 @@ impl<DB: Database> Client<DB> {
 
         let res = self.node.write().await.sync().await;
         if let Err(err) = res {
-            warn!("consensus error: {}", err);
+            warn!(
+                "failed to sync consensus node with checkpoint: {:?}. Error: {}",
+                self.node.read().await.config.checkpoint,
+                err
+            );
+            let fallback = if let Some(fallback) = &self.fallback {
```
ncitron, line 210: I think a better order of operations to do here is that if the initial sync fails:

1. if user has set a fallback url and has set the load from fallback flag, then use just that fallback url to fetch a checkpoint. If it fails, do not continue.
2. If user has not set a fallback url and has set the load from fallback flag, then use all the ones from repo to try.
3. If load from fallback is not set, always fail.

The reason I think this is useful, is that someone might want to set their fallback url in their config file, but never have that get used unless they have problems syncing. If they have a problem, they can use their prefered url to sync by just adding `-l`.

I also thing if someone explicitly sets a fallback url, they are sort of saying "I trust this one but not the others", so we shouldn't continue on to use the others.

Basically I think `-l` should act as a toggle and `-f` should act as a url override.

config/src/checkpoints.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Associated function to fetch the latest checkpoint from a specific checkpoint sync fallback
+    /// service api url.
+    pub async fn fetch_checkpoint_from_api(url: &str) -> Result<H256, Box<dyn std::error::Error>> {
+        // Fetch the url
+        let client = reqwest::Client::new();
+        let constructed_url = Self::construct_url(url);
+        let res = client.get(constructed_url).send().await?;
+        let raw: RawSlotResponse = res.json().await?;
+        let slot = raw.data.slots[0].clone();
```
ncitron, line 169: This can panic if the api returns an empty array

ncitron (commented):
looks really good

client/src/client.rs
ncitron: Actually the more I think about it I think the original way was better. What do you think?

config/src/checkpoints.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        // Parse the yaml content results.
+        let list: CheckpointFallbackList = serde_yaml::from_str(&yaml)?;
+        self.mainnet = list.mainnet;
+        self.goerli = list.goerli;
+        Ok(())
+    }
+
+    /// Fetch the latests mainnet checkpoint sync service checkpoint.
+    pub async fn fetch_latest_mainnet_checkpoint(
```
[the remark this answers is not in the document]
abigger87: Good call

ncitron (approved):
LGTM!

PR #124, opened 2022-12-02
fix: surface eth_getStorageAt

sragss:


ncitron:
Apparently `eth_getStorageAt` takes a block tag as well. Do you mind adding that in?

sragss:
Done.

ncitron (approved):
LGTM

PR #129, opened 2022-12-04
fix(readme): Resolve Dependency Type Conflicts

### LINKED ISSUE #127: mismatched types when attempting to import helios and ethers into the same project
Trying the sample code in the `README` leads me to a few compilation errors such as this:

```
error[E0308]: mismatched types
   --> /home/naps62/.cargo/git/checkouts/helios-b3bc464b79507b80/e4071fe/execution/src/evm.rs:256:13
    |
255 |         Ok(Some(AccountInfo::new(
    |                 ---------------- arguments to this function are incorrect
256 |             account.balance,
    |             ^^^^^^^^^^^^^^^ expected struct `primitive_types::U256`, found struct `ethers::types::U256`
    |
    = note: struct `ethers::types::U256` and struct `primitive_types::U256` have similar names, but are actually distinct types
```

I tried overriding `ethers` to use the same fork specified in this repo, but that seems to have changed nothing

refcell:
## Overview

Resolves dependency type conflicts between `revm` and `ethers`. Fixes #127

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,20 +25,20 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.7.6"
+version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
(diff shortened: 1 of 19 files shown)


ncitron (approved):
LGTM! Good to finally be one stable tags for both revm and ethers-rs, so we hopefully won't have this problem crop up anymore.

PR #131, opened 2022-12-06
feat(benches): Criterion + Iai Benchmarking [RFC]

refcell:
## Overview

Introduces criterion statistical benchmarking as well as iai one-shot cachegrind benchmarking.

Additionally, fixes a `Checkpoint` issue where unhealthy services prevent successful checkpoint fetches.

#### Benches

- `benches/file_db.rs`: benchmarks reading and writing to the `FileDB`
- `benches/get_balance.rs`: benchmarks fetching account balances on goerli and mainnet.
- `benches/get_code.rs`: benchmarks fetching address code on goerli and mainnet.

#### Checklist

- [x] Full Sync ("known" checkpoint)
- [x] Full Sync + Call ("known" checkpoint)
- [x] Full Sync (Checkpoint Fallback Service)
- [x] Full Sync + Call (Checkpoint Fallback Service)

_For "known" checkpoint, we fetch the latest checkpoint outside the bench_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,12 +6,15 @@ on:
   pull_request:
     branches: [ "master" ]
 
+env:
+  BENCHMARK_MAINNET_RPC_URL: ${{ secrets.BENCHMARK_MAINNET_RPC_URL }}
+  BENCHMARK_GOERLI_RPC_URL: ${{ secrets.BENCHMARK_MAINNET_RPC_URL }}
+
 jobs:
   check:
-    name: check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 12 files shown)


ncitron:
This is really cool. Didn't know about the criterion crate. 

We probably want to add the results of the benchmarks somewhere as well.

sragss:
Probably worth benchmarking at a lower level (skipping network) in addition to the integration benches. 

I would suggest a bench of a receipt proof + storage proof. Not a blocker on this PR though.

refcell:
> Probably worth benchmarking at a lower level (skipping network) in addition to the integration benches.
>
> I would suggest a bench of a receipt proof + storage proof. Not a blocker on this PR though.

Would you be willing to spec this out in an issue so I (or some other interested party) can implement in a diff pr?

ncitron:
I'm also not sure if we wan't to exclude the network from the bechmark. Although we have no control of it, the network is the slowest part of every call. I was thinking the benchmarks would just give a sense of how long Helios takes to do most things, rather than some theoretical "how fast is the code executing" metric.

refcell:
ye, taking offline so we can discuss best way forward

ncitron (approved):
LGTM

Issue #133, opened 2022-12-09
missing storage slots when getting accounts causing panics


ncitron:
Some calls are causing a very unexpected failure.

For example, calling `renderBroker(5)` on `0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac` always fails on line 287 of `evm.rs`. This is very unexpected, since in this case we are just fetching and proving that slot, and immediately fetching the slot out of the map and unwrapping it.

ncitron:
It gets even weirder. In this example, the createAccessList misses so we need to sequentially fetch a couple of slots. First we fetch some accounts slot 4b, and it works fine. Next we fetch that same accounts slot 80, yet the account returned by `execution` contains just the slot 4b.

PR #134, opened 2022-12-10
fix: Missed Storage Slots

refcell:
## Overview

Tracks #133 

#### Context

```markdown
Some calls are causing a very unexpected failure.

For example, calling `renderBroker(5)` on `0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac`
always fails on line 287 of `evm.rs`.
This is very unexpected, since in this case we are just fetching and proving that slot.
Then immediately fetching the slot out of the map and unwrapping it.
```

We can see the asset is successfully fetched using an access point: https://0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac.w3eth.io/renderBroker/5 as a sanity check.

#### Progress

Constructed an example in [examples/call.rs](./examples/call.rs) that calls the given method using both [ethers-rs provider](https://docs.rs/ethers/latest/ethers/providers/struct.Provider.html) as well as the Helios client.

Occasionally errors due to checkpoint sync issue it seems constructing access list when the client isn't fully synced yet.

See screengrabs below (both run within 1 minute of each other, no code changes).

Screenshot 2022-12-10 at 8 51 58 AM
Screenshot 2022-12-10 at 8 51 50 AM

### DIFF
--- a/execution/src/evm.rs
+++ b/execution/src/evm.rs
@@ -54,6 +54,8 @@ impl<'a, R: ExecutionRpc> Evm<'a, R> {
 
         self.evm.env = self.get_env(opts);
         let tx = self.evm.transact().0;
+        log::debug!("tx: {:#?}", tx);
+        log::debug!("tx.exit_reason: {:?}", tx.exit_reason);
 
         match tx.exit_reason {
             revm::Return::Revert => match tx.out {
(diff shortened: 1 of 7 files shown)


execution/src/evm.rs
```diff
@@ -277,21 +278,23 @@ impl<'a, R: ExecutionRpc> Database for ProofDB<'a, R> {
...
             Some(account) => match account.slots.get(&slot) {
-                Some(slot) => *slot,
-                None => *self
+                Some(slot) => Ok(*slot),
+                None => self
                     .get_account(address, &[slot])?
                     .slots
                     .get(&slot)
-                    .unwrap(),
+                    .cloned()
```
prestwich, line 288: nit: use `copied()` instead, as U256 is Copy

ncitron: Thanks for pointing this out. There are probably quite a few places where we can be better on copy vs clone.

ncitron:
closing in favor of #208

Issue #135, opened 2022-12-10
Missing RPC documentation


giovannivignone:
We are currently missing a central location to view all RPC methods implemented by Helios. The [rpc.rs file](https://github.com/a16z/helios/blob/master/client/src/rpc.rs) defines the methods but we do not have documentation.

ncitron:
closed by #136

PR #136, opened 2022-12-10
RPC Documentation

giovannivignone:
# Overview
PR for RPC documentation given issue #135 

## Changes
- Adding a rpc.md with a table of supported rpc methods.
- Linking to the rpc.md from readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,9 @@ helios --execution-rpc $ETH_RPC_URL
 
 Helios will now run a local RPC server at `http://127.0.0.1:8545`.
 
-Helios also provides examples in the [`examples/`](./examples/) directory. To run an example, you can execute `cargo run --example <example_name>` from inside the helios repository.
+Helios provides examples in the [`examples/`](./examples/) directory. To run an example, you can execute `cargo run --example <example_name>` from inside the helios repository.
 
+Helios also provides documentation of its supported RPC methods in the [rpc.md](./rpc.md) file.
 
 ### Warning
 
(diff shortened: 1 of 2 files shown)


ncitron (changes requested):
This looks mostly good to me. However, the rpc methods refers to the jsonrpc methods, not the client interface. For example, the client's `get_balance` is actually `eth_getBalance`. Can we use these instead? Also think its ok to remove the examples once we switch to this since the rpc is very well documented.

giovannivignone:
Yup no problem. @ncitron Basically change every left value in the table into the actual JSON RPC method, not the helios specific ones in rust and remove the examples? 

Just one question: shouldn't we display somewhere in this documentation how to actually call these methods with helios (like shouldn't we include the information that get_block() on the client is how you call eth_getBlock JSON RPC)?

refcell:
> This looks mostly good to me. However, the rpc methods refers to the jsonrpc methods, not the client interface. For example, the client's `get_balance` is actually `eth_getBalance`. Can we use these instead? Also think its ok to remove the examples once we switch to this since the rpc is very well documented.

+1 Maybe split the methods into two columns, something like:

```markdown
| RPC Method | Client Function |  Description | Example |
| ------------ | --------------- | ------------- | --------- |
```

ncitron:
Agreed @refcell

refcell (approved):
LGTM

ncitron (approved):
LGTM

PR #140, opened 2022-12-13
feat(docs): Config

refcell:
## Overview

Introduces comprehensive documentation of config file options.

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,0 +1,84 @@
+## Helios Configuration
+
+All configuration options can be set on a per-network level in `~/.helios/helios.toml`.
+
+#### Comprehensive Example
+
+```toml
+[mainnet]
+# The consensus rpc to use. This should be a trusted rpc endpoint. Defaults to "https://www.lightclientdata.org".
+consensus_rpc = "https://www.lightclientdata.org"
+# [REQUIRED] The execution rpc to use. This should be a trusted rpc endpoint.
+execution_rpc = "https://eth-mainnet.g.alchemy.com/v2/XXXXX"
+# The port to run the JSON-RPC server on. By default, Helios will use port 8545.
+rpc_port = 8545
+# The latest checkpoint. This should be a trusted checkpoint that is no greater than ~2 weeks old.
+# If you are unsure what checkpoint to use, you can skip this option and set either `load_external_fallback` or `fallback` values (described below) to fetch a checkpoint. Though this is not recommended and less secure.
+checkpoint = "0x85e6151a246e8fdba36db27a0c7678a575346272fe978c9281e13a8b26cdfa68"
+# The directory to store the checkpoint database in. If not provided, Helios will use "~/.helios/data/mainnet", where `mainnet` is the network.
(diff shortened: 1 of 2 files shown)


config.md
```diff
@@ -0,0 +1,84 @@
...
+rpc_port = 8545
+# The latest checkpoint. This should be a trusted checkpoint that is no greater than ~2 weeks old.
+# If you are unsure what checkpoint to use, you can skip this option and set either `load_external_fallback` or `fallback` values (described below) to fetch a checkpoint. Though this is not recommended and less secure.
+checkpoint = "0x85e6151a246e8fdba36db27a0c7678a575346272fe978c9281e13a8b26cdfa68"
+# The directory to store the checkpoint database in. If not provided, Helios will use "~/.helios/data/mainnet", where `mainnet` is the network.
+# It is recommended to set this directory to a persistent location mapped to a fast storage device.
+data_dir = "/home/user/.helios/mainnet"
+chain = "mainnet"
+# The forks to enable. This is a comma-separated list of fork names.
+forks = [ "berlin" ]
```
ncitron, line 23: Forks specify the consensus layer fork information and includes both the epoch and the fork version which is a unique identifier that the beaconchain uses in its signing operations to make signed messages unique to a specific fork. A fork name alone will not be parsed correctly.

We can probably just remove this, since it is very low level info that isn't really meant to be set by the user. Similar to a lot of the stuff in `ChainConfig`.

Here is the relevant type: https://github.com/a16z/helios/blob/master/config/src/types.rs#L24

config.md
```diff
@@ -0,0 +1,84 @@
...
+execution_rpc = "https://eth-mainnet.g.alchemy.com/v2/XXXXX"
+# The port to run the JSON-RPC server on. By default, Helios will use port 8545.
+rpc_port = 8545
+# The latest checkpoint. This should be a trusted checkpoint that is no greater than ~2 weeks old.
+# If you are unsure what checkpoint to use, you can skip this option and set either `load_external_fallback` or `fallback` values (described below) to fetch a checkpoint. Though this is not recommended and less secure.
+checkpoint = "0x85e6151a246e8fdba36db27a0c7678a575346272fe978c9281e13a8b26cdfa68"
+# The directory to store the checkpoint database in. If not provided, Helios will use "~/.helios/data/mainnet", where `mainnet` is the network.
+# It is recommended to set this directory to a persistent location mapped to a fast storage device.
+data_dir = "/home/user/.helios/mainnet"
+chain = "mainnet"
```
ncitron, line 21: The chain option actually refers to the `ChainConfig` struct. In theory you can include all this stuff with nested toml headers, but it might be better just to omit it since chain includes only low level options that should never be modified such as `genesis_validators_root`, `genesis_time`, and `chain_id`.

Here is the relevant type: https://github.com/a16z/helios/blob/master/config/src/types.rs#L6

ncitron (changes requested):

ncitron (approved):
LGTM

PR #143, opened 2022-12-14
fix(ci): move benchmark ci to seperate action

ncitron:
Right now, if someone opens up a PR from a fork, the benchmarks will fail. This is because actions triggered from PRs from forks cannot access secrets for security reasons.

One option is to use the `pull_request_target`  trigger, which allows PRs from forks to access secrets. I don't want to do this though in case we add more valuable secrets and forget that this action trigger can expose them to an attacker. Obviously as it stands, I don't really care if someone can steal my alchemy key though.

This PR just switches the benches to a separate action that only triggers on pushes to master. I don't think we often care about the benchmarks for every incoming PR anyway, and we can always manually trigger the benchmark action for a PR if we want.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,25 @@
+name: benchmarks
+
+on:
+  push:
+    branches: [ "master" ]
+
+env:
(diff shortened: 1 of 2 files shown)


refcell (approved):
Nice, this LGTM!

PR #146, opened 2022-12-14
chore: export ExecutionBlock type

msaug:
# Description
Currently execution::types::ExecutionBlock is not exported, and it is needed to be able to use `get_block_by_<x>` functions of Helios client.

## Changes summary
* Updated Helios lib.rs to export `execution::types::{CallOpts, ExecutionBlock}`

Closes https://github.com/a16z/helios/issues/145

PR #148, opened 2022-12-14
chore: export `src/execution` types

### LINKED ISSUE #147: Expose all types under `execution/src/types`
Following #145, additional types need to be exported in order to use the helios lightclient externally.

I will submit a PR proposing to export the remaining types, which will enable the use of the client functions from external locations.

msaug:
# Description
Following PR #146, other types need to be exported for external integration.
Currently `execution::types::Transactions` is not exported, and it is needed to be able to use `get_block_by_<x>` functions of Helios client.

To avoid future PRs solving the same problem, I propose to export all types in this file  - i.e. export `Account` and `Transactions`.

## Changes summary
* Updated Helios lib.rs to export `execution::types::{Account, CallOpts, ExecutionBlock, Transactions}`

Closes #147

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,7 +60,7 @@ pub mod config {
 
 pub mod types {
     pub use common::types::BlockTag;
-    pub use execution::types::{CallOpts, ExecutionBlock};
+    pub use execution::types::{Account, CallOpts, ExecutionBlock, Transactions};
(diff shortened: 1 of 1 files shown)


ncitron (approved):
LGTM. Thanks for going through and finding all the types we need to export.

PR #149, opened 2022-12-14
chore: add checkpoints in export config mod

danilowhk:
Added "checkpoints" to config module exports
I saw there were some more modules in config, so if there are more that needs to be added please let me know.

Issue: #150

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,7 @@ pub mod client {
 }
 
 pub mod config {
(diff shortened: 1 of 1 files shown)


ncitron (approved):
LGTM. 

With this you should have most anything that library users will care about. We might want to consider adding `types` and `base` in the future, but that stuff should never need to be changed anyway as its all very low level chain details.

Issue #151, opened 2022-12-14
Remaining RPC Methods


giovannivignone:
There are a few RPC methods from the [Ethereum RPC Spec](https://ethereum.github.io/execution-apis/api-documentation/) that have not been implemented within Helios yet. 

The main ones that might be useful to our users are the following...
1. eth_sync
2. eth_coinbase
3. eth_feeHistory
4. eth_newFilter
5. eth_newBlockFilter
6. eth_newPendingTransactionFilter
7. eth_uninstallFilter
8. eth_getFilterChanges
9. eth_getFilterLogs
10. eth_getTransactionByBlockHashAndIndex
11. debug_getRawHeader
12. debug_getRawBlock
13. debug_getRawTransaction
14. debug_getRawReceipts
15. debug_getBadBlocks

danilowhk:
Could I try to implement eth_getTransactionByBlockHashAndIndex?

giovannivignone:
@danilowhk yea go for it. I'm busy till Tuesday and then I'll get to the remaining methods. Just update here with progress through prs or comments :)

danilowhk:
Ok! :)

pradaaripp12:
can we implement with eth_getPendingTransaction

danilowhk:
@giovannivignone made the PR for eth_getTransactionByBlockHashAndIndex , please let me know if there are changes or improvements you would like me to add :)

giovannivignone:
@danilowhk nice work. This looks great. I left one comment, feel free to reply and let me know what you think. Thanks for getting this in.

dadepo:
@giovannivignone thinking of helping out with these rpc methods. Looks like `eth_feeHistory` will be quick one to knock out. if you or anyone else is not working on this, I'll like to pick it up

giovannivignone:
@dadepo go for it. Let me know if you need any help

danilowhk:
@giovannivignone I would like to implement eth_coinbase and eth_newFilter if its ok :)

giovannivignone:
@danilowhk go for it

PR #157, opened 2022-12-20
RPC call : eth_getTransactionByBlockHashAndIndex Implementation

danilowhk:
- Add get_transaction_by_block_hash_and_index to client
- Add the same function in node, which calls execution's version of it (get_transaction_by_block_hash_and_index) 
 OBS: Decided to implement on execution level and use block data from payload and get_transaction from rpc's
- Implement this new RPC method on RpcInner (the server) by calling into Client::getTransactionByBlockHashAndIndex()
- Added a test for this functionality in execution/tests/execution.rs

Issue: #151

### DIFF
--- a/execution/src/execution.rs
+++ b/execution/src/execution.rs
@@ -184,18 +184,15 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
         payload: &ExecutionPayload,
         index: usize,
     ) -> Result<Option<Transaction>> {
-
-
         let tx = payload.transactions[index].clone();
         let tx_hash = H256::from_slice(&keccak256(tx));
         let mut payloads = BTreeMap::new();
         payloads.insert(payload.block_number, payload.clone());
         let tx_option = self.get_transaction(&tx_hash, &payloads).await?;
         let tx = tx_option.ok_or(eyre::eyre!("not reachable"))?;
-    
-        Ok(Some(tx))
 
-    }    
+        Ok(Some(tx))
+    }
 
     pub async fn get_transaction_receipt(
         &self,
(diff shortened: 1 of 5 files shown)


execution/src/execution.rs
```diff
@@ -179,6 +179,21 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
...
+    pub async fn get_transaction_by_block_hash_and_index(
+        &self,
+        payload: &ExecutionPayload,
+        index: usize,
+    ) -> Result<Option<Transaction>> {
+        let tx = payload.transactions[index].clone();
+        let tx_hash = H256::from_slice(&keccak256(tx));
+        let mut payloads = BTreeMap::new();
+        payloads.insert(payload.block_number, payload.clone());
+        let tx_option = self.get_transaction(&tx_hash, &payloads).await?;
```
giovannivignone, line 191: Here I think we should use the rpc method you built: self.rpc.get_transaction_by_block_hash_and_index rather than self.get_transaction

danilowhk: Ok! I was thinking of which one to use and chose this one(rpc level or execution level) but will implement the Rpc level implementation then

ncitron: Just make sure if you do choose to use the rpc, you need to be very careful to verify that the returned tx has the same hash as the hash at that location in the payload. Otherwise it would be possible for the rpc to send a fake tx back.

danilowhk: I see, should we keep the current implementation then? As from what I understood if we use the rpc call, we will basically need to access "almost" the same data we are accessing now and then compare

danilowhk: Or would be nice to have the comparison(fetching from both payload and rpc) as it will be an extra layer of security?

ncitron: I think this is good. Sorry for the late response here, missed the notification.

ncitron (approved):
LGTM

PR #160, opened 2022-12-23
Reduce Binary Size to 5.3mb

ControlCplusControlV:
Using some compiler options I managed to reduce the binary to 5.3MB. I consider the tradeoff for helios strictly in favor of size, as it's greatly increases the accessibility of the light client for it to be embedded in webpages and such. 

With the nightly feature `panic_immediate_abort` the binary size can be further reduced and cut out some more panic string formatting. This provides a nice baseline for further size optimizations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,13 @@ plotters = "0.3.3"
 tempfile = "3.3.0"
 hex = "0.4.3"
 
+[profile.release]
+strip = true
+opt-level = "z"
+lto = true
+codegen-units = 1
+panic = "abort"
+
 ######################################
 # Examples
 ######################################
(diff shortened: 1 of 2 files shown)


ncitron (approved):
LGTM! Was cool to read about some of these compiler options.

cawfree:
Hey y'all, just a heads up that the option `panic = "abort"` appeared to regress the iOS builds on [`react-native-helios`](https://github.com/cawfree/react-native-helios) with the following error:

```
duplicate lang item in crate `core`: `sized`
```

This had no effect upon the Android build process. I've disabled this setting upstream.

ControlCplusControlV:
To save on Android builds if you know the full target name it could be set as a specific setting for the IOS target

cawfree:
@ControlCplusControlV Thanks for your help! Luckily the react-native port depends on a pretty unconventional build process for each platform, [so we were able to remove `panic = "abort"` just for iOS but still hold onto it for Android](https://github.com/cawfree/react-native-helios/blob/64029d149eda3c6cb329ab129f64939b03cb6ac7/scripts/heliosup.ts#L652). I'm assuming it has something to do with the `ffi` bridge.

Issue #141, opened 2022-12-13
add interaction diagram


Eikix:
Hey y'all! Found your project through [beerus](https://github.com/keep-starknet-strange/beerus), a Starknet light client that leverages helios. 

I'm trying to understand the interaction flow that occurs in a helios light client and thought that a simple diagram (excalidraw-like) could help! 

something in the spirit of 

![image](https://user-images.githubusercontent.com/66871571/207294884-58c99daf-126f-4609-bae2-13189ae99584.png)

![image](https://user-images.githubusercontent.com/66871571/207294911-0caf5fa0-9ef0-426b-9b84-4b5c22394db9.png)

Thank you all for your help!

giovannivignone:
I think this would be great. I was going to try to do this using mermaid. I would recommend using mermaid rather than excalidraw for the sake of easy future modification

ncitron:
@giovannivignone is this something you would be interested in tackling?

giovannivignone:
@ncitron Not sure if I have enough context yet on the entire repo to do this accurately. But if you want to design something in excalidraw or provide some detailed notes, I can reproduce it in mermaid. If it is easier to just tackle this yourself, im happy to review. Definitely something we should have though.

Eikix:
Fair point about mermaid! Even saw it possible to import as markdown format!

giovannivignone:
@Eikix @ncitron Let me know if this diagram fits the architecture. Easiest to view [here](https://github.com/giovannivignone/helios#architecture)

Eikix:
Ggs people:)! I'm not sure I understand the flow though

When someone goes through the Helios client, the helios client calls an helios node, that in turn calls an execution node and a consensus node, which give access to a trustless execution rpc invoke and trustless consensus rpc invoke?

I'm not sure I get the flow

ncitron:
You've got he basic flow right, but each part serves a specific purpose. The call to `ConsensusClient` is able to retrieve provable block headers (using the eth2 light client protocol) from the untrusted consensus node. From there `Node` takes that data, and sends it to `ExecutionClient`. `ExecutionClient` is then able to talk to an untrusted execution node and check the data it receives from that against the data from the consensus side.

For specifics on how this all works, I'd suggest taking a look at [our blog post](https://a16zcrypto.com/building-helios-ethereum-light-client/).

Eikix:
[...]

PR #161, opened 2022-12-25
Architectural Diagram

giovannivignone:
# Diagram for Helios Architecture
- As requested in telegram and in #141, @ncitron and I have produced an architectural overview of Helios
- This architecture is up for review and anyone can feel free to leave comments.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,25 @@
+name: benchmarks
+
+on:
+  push:
+    branches: [ "master" ]
+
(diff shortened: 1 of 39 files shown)


ncitron (commented):
Can we show that there is two untrusted RPC providers (consensus layer and execution layer) rather than just a single one that the diagram currently shows? Other than that everything else looks great.

giovannivignone:
@ncitron Done. Let me know what you think!

ncitron:
Looks great!

PR #166, opened 2023-01-11
feat : add eth_getTransactionCount implementation

ftupas:
Hello, this pr implements the following

-  Add get_transaction_count to client
-  Add the same function in node, which calls ExecutionClient
-  Implement this RPC method on ExecutionRPC which calls the provider that sends a request to the `eth_getTransactionCount` endpoint
-  Added a test for this functionality in execution/tests/execution.rs

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -398,6 +398,14 @@ impl<DB: Database> Client<DB> {
         self.node.read().await.get_logs(filter).await
     }
 
+    pub async fn get_transaction_count(&self, address: &Address, block: BlockTag) -> Result<U256> {
+        self.node
+            .read()
+            .await
+            .get_transaction_count(address, block)
+            .await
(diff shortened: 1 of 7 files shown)


ncitron:
We already have this implemented in `Client` and `Node` it is just called `get_nonce`. We're planning to rename in on the next major version change, but don't want to introduce breaking changes so we are holding off. The rpc method is also implemented in `client/rpc.rs`.

Appreciate the effort though.

ftupas:
> We already have this implemented in `Client` and `Node` it is just called `get_nonce`. We're planning to rename in on the next major version change, but don't want to introduce breaking changes so we are holding off. The rpc method is also implemented in `client/rpc.rs`.
> 
> Appreciate the effort though.

Ah got it, no worries! Thank you @ncitron  🙏

PR #167, opened 2023-01-11
RPC call : eth_coinbase Implementation #157

danilowhk:
- Implemented get_coinbae Rpc method

### DIFF
--- a/client/src/node.rs
+++ b/client/src/node.rs
@@ -10,7 +10,7 @@ use common::errors::BlockNotFoundError;
 use common::types::BlockTag;
 use config::Config;
 use consensus::rpc::nimbus_rpc::NimbusRpc;
-use consensus::types::{ExecutionPayload, Header};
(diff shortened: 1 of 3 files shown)


client/src/node.rs
```diff
@@ -291,6 +291,13 @@ impl Node {
         Ok(self.consensus.get_header().clone())
     }
 
+    pub fn get_coinbase(&self) -> Result<CoinbaseAddress> {
```
ncitron, line 294: Can we just have it return an ethers-rs `Address`? The consensus type is more just for internal use and has some special serializations. Trying to use ethers types as the outputs when possible as most people are familiar with that. Also means we won't need to export a new type.

Same goes for in `Client` and `Rpc`.

Other than that this all looks good.

danilowhk: Ok!

danilowhk: Done, please let me know if the implementation is ok

ncitron (approved):
LGTM!

PR #179, opened 2023-01-24
fix: specify precise nightly version

prestwich:
Today seems as good a day as any. This will cause reinstallation of the nightly tool chain, and cache misses on first build. I have built locally and run cargo fmt using this toolchain file.


ncitron (approved):
LGTM! Thanks for the contribution.

emlazzarin:
Thanks @prestwich!

Issue #163, opened 2023-01-09
backfill payloads at start


ncitron:
When Helios starts up, it only tracks blocks that are added to the chain after the startup time. This means that if someone requests data from a blocktag of just a few blocks ago near startup, then Helios will error. While this is originally the expected behavior, it would be nice not to have this problem. When solution is simply to backfill payloads from the start. Once we have the current payload, just request the previous `HISTORY_LENGTH` payloads and check the previous block hash field on each one to ensure its validity.

PR #189, opened 2023-02-03
feat: backfill payloads

ckoopmann:
resolves: https://github.com/a16z/helios/issues/163

Additionally backfills payloads since the last update to avoid gaps in the range of available blocks.

Note that the only check done on the backfilled blocks, is verifying the "parent block hash chain" starting from the "optimistic_header". Not sure if this is sufficient.

### DIFF
--- a/client/src/node.rs
+++ b/client/src/node.rs
@@ -5,14 +5,15 @@ use std::time::Duration;
 use ethers::prelude::{Address, U256};
 use ethers::types::{Filter, Log, Transaction, TransactionReceipt, H256};
 use eyre::{eyre, Result};
+use futures::future::join_all;
 
 use common::errors::BlockNotFoundError;
 use common::types::{BlockTag, Bytes32};
 use config::Config;
+use consensus::errors::ConsensusError;
 use consensus::rpc::nimbus_rpc::NimbusRpc;
 use consensus::types::{ExecutionPayload, Header};
 use consensus::ConsensusClient;
-use consensus::errors::ConsensusError;
 use execution::evm::Evm;
 use execution::rpc::http_rpc::HttpRpc;
 use execution::types::{CallOpts, ExecutionBlock};
@@ -93,29 +94,17 @@ impl Node {
     }
 
     async fn update_payloads(&mut self) -> Result<(), NodeError> {
-        let latest_header = self.consensus.get_header();
-        let start_slot = self.current_slot.unwrap_or(latest_header.slot);
+        let latest_header = self.consensus.get_header().clone();
+        let start_slot = self
+            .current_slot
+            .unwrap_or(latest_header.slot - self.history_size as u64);
         info!(
             "updating payloads   payloads_before={:?}",
             self.payloads.keys(),
         );
 
-        let mut prev_parent_hash: Option<Bytes32> = None;
-        for slot in (start_slot..=latest_header.slot).rev() {
-            let payload = self
-                .consensus
-                .get_execution_payload(&Some(slot))
-                .await
-                .map_err(NodeError::ConsensusPayloadError)?;
-            if let Some(parent_block_hash) = &prev_parent_hash {
-                if &payload.block_hash != parent_block_hash {
-                    return Err(NodeError::ConsensusPayloadError(ConsensusError::InvalidHeaderHash(
-                            format!("{:02X?}", parent_block_hash.to_vec()),
-                            format!("{:02X?}", payload.parent_hash.to_vec())
-                    ).into()));
-                } 
-            }
-            prev_parent_hash = Some(payload.parent_hash.clone());
+        let payloads = self.get_payloads(start_slot, latest_header.slot).await?;
+        for payload in payloads {
             self.payloads.insert(payload.block_number, payload);
         }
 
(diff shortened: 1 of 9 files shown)


client/src/node.rs
```diff
@@ -123,6 +136,49 @@ impl Node {
...
+                    .execution_payload;
+                Ok::<ExecutionPayload, NodeError>(payload)
+            };
+            inner_closure.await
+        });
+        let mut prev_parent_hash: Option<Bytes32> = None;
+        let mut payloads: Vec<ExecutionPayload> = Vec::new();
+        for result in join_all(payloads_fut).await {
+            if result.is_err() {
+                // If fetching any payload failed we only backfill up to that payload
```
ckoopmann, line 158: Note that this can result in "holes" in the backfilled blocks, because at the next attempt backfilling will only start at the latest block from the previous update.

An alternative here would be to check all slots since `history_size` blocks ago in the `self.payloads` mapping and attempting to backfill any gaps. I can refactor it accordingly if it's worth it.

client/src/node.rs
```diff
@@ -123,6 +136,49 @@ impl Node {
         Ok(())
     }
 
+    async fn get_payloads(&self, start_slot: u64, end_slot: u64) -> Result<Vec<ExecutionPayload>, NodeError> {
+        let payloads_fut = (start_slot..end_slot).rev().into_iter().map(|slot| async move {
```
ckoopmann, line 140: I made this concurrent to speed up the backfill, not sure if this might lead to rate limit failures on any public beacon api though and if it is worth adding some kind of "retry" logic for that case.

ncitron: I wouldn't worry about that for now. Most people use the default one I set, which doesn't rate limit. We can address the problem if it arises in the future.

client/src/node.rs
```diff
@@ -123,6 +136,49 @@ impl Node {
         Ok(())
     }
 
+    async fn get_payloads(&self, start_slot: u64, end_slot: u64) -> Result<Vec<ExecutionPayload>, NodeError> {
```
ncitron, line 139: Can we move this into `ConsensusClient`? Want to have most consensus related stuff separated, so that we can support over evm chains or L2s in the future by swapping out to a new `ConsensusClient`.

client/src/node.rs
```diff
@@ -123,6 +136,49 @@ impl Node {
...
+        let mut payloads: Vec<ExecutionPayload> = Vec::new();
+        for result in join_all(payloads_fut).await {
+            if result.is_err() {
+                // If fetching any payload failed we only backfill up to that payload
+                // TODO: Review if it is better to error out of the whole payload update instead;
+                break;
+            }
+            let payload = result?;
+            if let Some(parent_block_hash) = &prev_parent_hash {
+                if &payload.block_hash != parent_block_hash {
```
ncitron, line 164: I think this will break if a slot is missed at any point in the chain. I'm not exactly sure what `rpc.get_block` returns in the case of a missed slot, but we should find out and handle that case.

client/src/node.rs
```diff
@@ -123,6 +136,49 @@ impl Node {
...
+                Ok::<ExecutionPayload, NodeError>(payload)
+            };
+            inner_closure.await
+        });
+        let mut prev_parent_hash: Option<Bytes32> = None;
+        let mut payloads: Vec<ExecutionPayload> = Vec::new();
+        for result in join_all(payloads_fut).await {
+            if result.is_err() {
+                // If fetching any payload failed we only backfill up to that payload
+                // TODO: Review if it is better to error out of the whole payload update instead;
```
ncitron, line 159: I think the current handling is fine, no need to fully error out.

ncitron (commented):
Looking good. Just left a couple of minor comments.

Good to see another indexcoop friend in here!

client/src/node.rs
ckoopmann: Sure, good point. I moved the code over accordingly.

client/src/node.rs
ckoopmann: Good point. I don't know what the behaviour would be in that case either. Will try to find a skipped block and find out

ckoopmann: Another way we could do it would be just sequentially requesting the next block using the current `parent_block_hash`, which would probably fix this issue, but in that case we couldn't do the rpc requests concurrently.

ckoopmann: Looks like the consensus rpc returns a 404 Error when requesting a skipped block:
image

ckoopmann: This means that in the current implementation it will stop backfilling at that point due to the break statement [here](          https://github.com/a16z/helios/blob/0822f5874a759d6673be31c74d43417adf25acb6/consensus/src/consensus.rs#L123)

ckoopmann: I think this is the source of some intermittent errors that I had observed and which I assumed to be due to rate limiting.

ckoopmann: The easiest way to (hopefully) fix this is to just continue with the next block in case any rpc request failed. 
This should allow it to continue working if there is a skipped block, although the effectively backfilled history would be 1 element shorter than `self.history_size` in that case, which should not be too much of an issue I guess.

See: 
https://github.com/a16z/helios/pull/189/commits/48499addbb84bff58b9461376554ead47fcf35d3

ckoopmann:
> Looking good. Just left a couple of minor comments.
> 
> Good to see another indexcoop friend in here!

Pleasure's all mine. 
I went ahead and applied your suggestions. 

I only did "manual" testing on my machine. Lmk if I should add automated testing to the ci for this feature.

ckoopmann:
> Looks like clippy is still acting up. Can you resolve those? Also can you remove the `get_block_from_rpc` method since it is sort of redundant now?
> 
> Once thats done, will merge.

Clippy should be happy now. 

Btw:
I noticed that I had to run clippy on the stable toolchain locally (rather than nightly) in order to reproduce the CI result. 
(i.e. `cargo clippy --all -- -D warnings`). When I ran it with the `+nightly` tag I got a different set of warnings / errors, which is strange since the CI looks to be configured to use the nightly toolchain as well.  🤔

ncitron (approved):
LGTM!

PR #201, opened 2023-03-04
Consensus p2p tracking

GeemoCandama:
This document describes a couple of options for integrating consensus p2p laid out by refcell and ralexstokes: 
https://hackmd.io/BihePpvfQ3mSH2MZiVBT1Q

Option 1: Use mostly the same interface. With this we would just need the RPC to run in a background thread managing peers and caching messages, and ConsensusClient could just interact with it via a ConsensusRpc trait impl. This requires less changes to the surrounding consensus package.

Because this requires less changes I think this will be a bit easier.

refcell had a good plan in their PR so I'll just copy it here:
Specs (h/t @ralexstokes)

    HackMD: https://hackmd.io/@helios-core/p2p/edit

Implementation

- Setup: P2P Feature flag
- Step 1: Peer Discovery
- Step 2: Subscribe to light client topics with gossipsub protocol
- Step 4: Peer Manager
- Step 5: Adapt topic responses to existing light client data structures

- Testing
- Documentation

Currently I've started to integrate the peer discovery protocol, but it doesn't work as an integrated system yet. I'm looking for a bit of feedback on how this is integrating with the existing codebase.

Overview of current additions:

- Change the ConsensusRpc trait name to ConsensusNetworkInterface. (A change was suggested in the doc let me know what you think of the name)
- add a cli flag for p2p
- add a crate feature which conditionally compiles the additional dependencies for p2p
- made Node, Rpc, RpcInner generic over the Consensus Network Interface
- made ConsensusNetworkInterface::new async so it can start the discovery process and you could add a feature to fetch the rpc url/path from a cached file later if you want. (Not convinced this is the best approach. I'm thinking now it should probably start discovery during Client::start but idrk.
- add discovery example that really isn't doing much at this point 
note: there are warnings for unused code that will be used when not using hardcoded values

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,14 +19,116 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "aead"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7fc95d1bdb8e6666b2b217308eeeb09f2d6728d104be3e31916cc74d15420331"
+dependencies = [
+ "generic-array 0.14.6",
+]
+
+[[package]]
+name = "aead"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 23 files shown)


GeemoCandama:
I'm closing this as I don't have time to work on this anymore.

Issue #200, opened 2023-03-03
Missing RPC methods


ncitron:
Missing RPC methods that might be useful to have.

Feel free to reach out here or on the telegram if you want to discuss how to go about implementing any of these.

- [x] eth_sync
- [x] eth_coinbase
- [ ] eth_feeHistory
- [x] eth_newFilter
- [x] eth_newBlockFilter
- [x] eth_newPendingTransactionFilter
- [x] eth_uninstallFilter
- [x] eth_getFilterChanges
- [ ] eth_getFilterLogs
- [x] eth_getTransactionByBlockHashAndIndex
- [ ] debug_getRawHeader
- [ ] debug_getRawBlock
- [ ] debug_getRawTransaction
- [ ] debug_getRawReceipts
- [ ] debug_getBadBlocks

sk1122:
@ncitron what do you think?

sk1122:
> That's a good point @sk1122. Is it possible to check filter validity ourselves?

No, as they are generated [randomly](https://github.com/ethereum/go-ethereum/blob/master/rpc/subscription.go#L48), there is no way to verify them based on filter data 
 
> I don't think it's a huge deal anyway though, as for the log methods we sort of assume that it is possible for the rpc to withhold logs from us.

Understood, I think we should check it once before handing filter id over to user, what do you think?

Also in future, we can implement DA proofs and Fraud proofs to better secure us against these type of errors?

ncitron:
> Understood, I think we should check it once before handing filter id over to user, what do you think?

How would we check it before handing the id to the user?

sk1122:
we can check if `eth_getFilterChanges` returns an error, if it returns an error meaning that `id` wasn't registered on nodes (execution node can withheld that data though), else means that `id` has been successfully registered and now can be used for tracking changes

sk1122:
but for this, we will need to implement `eth_getFilterChanges`, which returns list of logs if Id is valid, we can just use `eth_getLogs` verification code for verifying this logs

sk1122:
so, should I first open a PR for eth_getFilterChanges?

ncitron:
@sk1122 that would be great!

muang0:
@ncitron I'm interested in contributing one of these methods, any preference which one I should do? Or is there another issue you think would be better suited for a 2nd contrib?

ncitron:
[...]

PR #204, opened 2023-03-09
eth_feeHistory Implementation

SFYLL:
Implement eth_feeHistory as requested in [#200](https://github.com/a16z/helios/issues/200) .

Veracity check on public RPC response done by comparing the response's base fee per gas with the local view provided by the `ExecutionPayload` object. This approach has two main corollaries:

- We take an optimistic view by assuming that if the response's base fee per block are correct, so are the rewards and gas ratios. Obviously, that's a subject for conversation;
- The `get_fee_history` in `execution/src/execution.rs`function checks the parameters and discards blocks that are not stored (and as such "validated") in `ExecutionPayload`. That makes returned values slightly less deterministic than one might have hoped, but seem in line with [Alchemy's](https://docs.alchemy.com/reference/eth-feehistory).

Given that the most insightful test were done by impersonating the user, i.e. putting in place integration tests, these were implemented in a new directory `tests/`.

Please note that this is my first PR into a public project + first time coding in Rust. So feel free to go heavy on the changes !


ncitron:
This is a good start. I do however think that we shouldn't trust the rest of response given that just part of it is provably correct. You can imagine attacks where an RPC sends correct base fees, but wildly high priority fees to trick Helios users into sending transactions with a high priority fee. They could then auction that artificially valuable transaction off to a miner using something like flashbots.

We should however be able to implement this method without needing any trust at all. Since Helios has past`ExecutionPayload`s, we can already get the base fees that we need. We also have the transaction list as a member of each `ExecutionPayload`, which we can use to decode the priority fees paid out in that block. With that list of priority fees, and a bit of math to figure out percentiles, we should be able to return a response with zero trust in the rpc.

Does this strategy make sense?

execution/src/execution.rs
```diff
@@ -345,6 +346,102 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
...
+
+    pub async fn get_fee_history(
+        &self,
+        block_count: u64,
+        last_block: u64,
+        reward_percentiles: &[f64],
+        payloads: &BTreeMap<u64, ExecutionPayload>,
+    ) -> Result<Option<FeeHistory>> {
+        let helios_latest_block = *payloads.last_key_value().unwrap().0;
+        let helios_oldest_block = *payloads.first_key_value().unwrap().0;
```
refcell, line 358: Unwraps will cause the program to panic (abort in release mode) - we should gracefully return an error here instead.

Example:

```suggestion
        let helios_latest_block = *payloads.last_key_value();
        let helios_latest_block = helios_latest_block.ok_or(BlockNotFoundError::new(BlockTag::Latest))?.0;
        let helios_oldest_block = *payloads.first_key_value();
        let helios_oldest_block = helios_oldest_block.ok_or(BlockNotFoundError::new(BlockTag::Number(0)))?.0;
```

(I would also allow `BlockNotFoundError` to accept an optional `BlockTag` so we can construct an error without knowing the actual block)

execution/src/execution.rs
```diff
@@ -345,6 +346,102 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
...
+        );
+
+        //case: requested block is more recent than helios' newest block
+        if request_latest_block > helios_latest_block {
+            request_latest_block = helios_latest_block;
+        }
+
+        //Can't prove anything in the request range
+        if request_latest_block < helios_oldest_block {
+            return Ok(None);
```
refcell, line 380: I would return an error here personally - the end consumer should know if their request is out of the block range.

execution/src/execution.rs
```diff
@@ -345,6 +346,102 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
...
+            //start from oldest block and validate the gas fee returned using helios' view
+            let block_to_check = (fee_history.oldest_block + _pos as u64).as_u64();
+            let payload = payloads.get(&block_to_check);
+
+            //Because of filtering done earlier in the function, there should be no block that we don't find in the payload
+            if payload.is_none() {
+                return Ok(None);
+            }
+
+            let payload = payload.unwrap();
```
refcell, line 411: ibid. Gracefully handle error/panic here.

SFYLL:
Hey guys,

Tried my best but unfortunately computing the reward parameters ourselves has quite some overhead (can push on another branch if needed, just let me know). As can be seen with [reth](https://github.com/paradigmxyz/reth/blob/ea11461126229a75532bc96d97e615d3b47ea343/crates/rpc/rpc/src/eth/api/server.rs#LL290C33-L290C34), some client actually don't give an answer for this parameter. 

I tried though, and unfortunately amount of request quickly starts going out of hand. Indeed, you need the gas_used parameter per transaction to accumulate it until you reach the gas_used percentile, at which point you can return the gas paid for that specific transaction. Ultimately, what that means is that you need to get the receipt for every transaction in each block queried as there is no other way to get the gas used (practical way at least, not planning to track opcode raised for that endpoint 😬 ) ...

If there was a way to prove that the content was correct without querying it ourselves (wink wink zero-knowledge proof), it would make getting that list possible. Maybe we can follow reth for now and re-assess later on?

refcell (changes requested):
Some minor changes requested. Overall, very nice work!

execution/src/errors.rs
```diff
@@ -26,6 +26,16 @@ pub enum ExecutionError {
...
     #[error("execution rpc is for the incorect network")]
     IncorrectRpcNetwork(),
+    #[error("Invalid base gas fee helios {0} vs rpc endpoint {1} at block {2}")]
+    InvalidBaseGaseFee(U256, U256, u64),
+    #[error("Invalid gas used ratio of helios {0} vs rpc endpoint {1} at block {2}")]
+    InvalidGasUsedRatio(f64, f64, u64),
+    #[error("Block {0} not found")]
+    BlockNotFoundError(u64),
+    #[error("Helios Execution Payload is empty")]
+    EmptyExecutionPayload(),
```
refcell, line 36: nice!

refcell (approved):
Very nice!

refcell:
Hey @SFYLL, merging into a fee_history branch in the `a16z/helios` repo so we can resolve these formatting and api key issues locally :)

Issue #153, opened 2022-12-15
Panic


eskimo:
Finally got it running only to now get this output:

```
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /home/eth/helios/execution/src/evm.rs:283:26
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /home/eth/helios/execution/src/evm.rs:283:26
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /home/eth/helios/execution/src/evm.rs:283:26
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /home/eth/helios/execution/src/evm.rs:283:26
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /home/eth/helios/execution/src/evm.rs:283:26
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /home/eth/helios/execution/src/evm.rs:283:26
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /home/eth/helios/execution/src/evm.rs:283:26
```

ncitron:
Do you happen to know what call your rpc is making when it crashes? This looks like it may be a bug which we've seen happen intermittently, so we'd appreciate knowing all the ways to trigger it.

We're tracking that issue at #133

cc @refcell

eskimo:
> Do you happen to know what call your rpc is making when it crashes? This looks like it may be a bug which we've seen happen intermittently, so we'd appreciate knowing all the ways to trigger it.
> 
> We're tracking that issue at #133
> 
> cc @refcell

Unfortunately I do not. These calls are being made by https://github.com/lukeburns/hipr-eth if that helps?

refcell:
Fixed in #208

PR #208, opened 2023-03-10
fix(execution/evm): evm panic on slot not found

refcell:
**Overview**

Attempts to address #153 

Inside a `storage` call to the `ProofDB`, unwraps were being performed for slot queries. This pr introduces a new error type to gracefully bubble up failed storage reads: `SlotNotFoundError`.


ncitron (approved):
LGTM

PR #209, opened 2023-03-10
feat: fee history

refcell:
**Overview**

Exposes logic to get the fee history. Extends #204 

Below written by @SFYLL in #204 🙏 

Implement eth_feeHistory as requested in [#200](https://github.com/a16z/helios/issues/200) .

Veracity check on public RPC response done by comparing the response's base fee per gas with the local view provided by the `ExecutionPayload` object. This approach has two main corollaries:

- We take an optimistic view by assuming that if the response's base fee per block are correct, so are the rewards and gas ratios. Obviously, that's a subject for conversation;
- The `get_fee_history` in `execution/src/execution.rs`function checks the parameters and discards blocks that are not stored (and as such "validated") in `ExecutionPayload`. That makes returned values slightly less deterministic than one might have hoped, but seem in line with [Alchemy's](https://docs.alchemy.com/reference/eth-feehistory).

Given that the most insightful test were done by impersonating the user, i.e. putting in place integration tests, these were implemented in a new directory `tests/`.


tests/feehistory.rs
```diff
@@ -0,0 +1,125 @@
...
+use eyre::Result;
+use helios::{config::networks::Network, prelude::*};
+use std::{env, path::PathBuf};
+
+#[tokio::test]
+async fn feehistory() -> Result<()> {
+    //Instantiate Client
+    env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
+
+    let api_key = env::var("YOUR_API_KEY").expect("Missing API key in environment variables, just run export YOUR_API_KEY='your-api-key-goes-here'");
```
ncitron: We already have an rpc url in our secrets called `MAINNET_RPC_URL`. If we switch the `untrusted_rpc_url` to that it should work.

execution/src/execution.rs
```diff
@@ -345,6 +346,108 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
...
+
+        let request_oldest_block = request_latest_block - block_count;
+        //case: request oldest block is further out than what helios' is aware of
+        if request_oldest_block < helios_oldest_block {
+            block_count = request_latest_block - helios_oldest_block + 1;
+        }
+
+        let fee_history = self
+            .rpc
+            .get_fee_history(block_count, request_latest_block, reward_percentiles)
```
ncitron, line 401: Do we actually need to call the rpc here? As we discussed, I think we were planning on omitting the reward percentiles since we cannot verify them. In that case, the remaining fields can all be fully computed (as you do here). Rather than computing and checking for correctness, we can just use the computed values.

ncitron (commented):
Mostly looking good. Two small points

tests/feehistory.rs
refcell: Switched

ncitron (approved):
LGTM

ncitron:
can u fmt real quick before merging?

PR #224, opened 2023-04-10
feat: save checkpoint after sync

glihm:
The last checkpoint of the node is only saved on `shutdown` from what I have seen.
But for some applications, like Beerus for instance, we are not always stopped gracefully. Which makes us always sync and we never have checkpoint saved into the `data_dir`.

In this PR, I refactorized the save checkpoint logic in the client into a function `save_last_checkpoint`. This function is called in `shutdown` as before, and after the sync with the node. I am not sure about the side effect of this, but this is working fine for us at this moment.

I kept this function private, but there are side effects I don't see and you do, I can propose:

1. Remove the call to `save_last_checkpoint` I made.
2. Make `save_last_checkpoint` public to let the library user decides when to save the checkpoint? Or there is too much internal logic involved, and we will then consider applying hook on signals to have `shutdown` called correctly anyway.

Thank you very much, and please let me know if some side effects are hidden, to better understand what's going on.

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -397,9 +397,10 @@ impl<DB: Database> Client<DB> {
     /// Saves last checkpoint of the node.
     async fn save_last_checkpoint(&self) {
         let node = self.node.read().await;
-        let _checkpoint = if let Some(_checkpoint) = node.get_last_checkpoint() {
+
+        if let Some(checkpoint) = node.get_last_checkpoint() {
             info!("saving last checkpoint hash");
-            let res = self.db.save_checkpoint(_checkpoint);
+            let res = self.db.save_checkpoint(checkpoint);
             if res.is_err() {
                 warn!("checkpoint save failed");
             }
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -286,6 +286,8 @@ impl<DB: Database> Client<DB> {
             }
         }
 
+        self.save_last_checkpoint().await;
+
         self.start_advance_thread();
 
         Ok(())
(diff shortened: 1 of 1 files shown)


client/src/client.rs
```diff
@@ -392,19 +394,20 @@ impl<DB: Database> Client<DB> {
...
 
-    pub async fn shutdown(&self) {
+    /// Saves last checkpoint of the node.
+    async fn save_last_checkpoint(&self) {
         let node = self.node.read().await;
-        let checkpoint = if let Some(checkpoint) = node.get_last_checkpoint() {
-            checkpoint
-        } else {
-            return;
+        let _checkpoint = if let Some(_checkpoint) = node.get_last_checkpoint() {
```
ncitron, line 400: Why are we declaring this first `_checkpoint` variable here?

Also can we rename the second `_checkpoint` variable to `checkpoint`. The underscore prefix is generally reserved for unused variables.

glihm: Sorry no reason, I tested some code to compile before and I had a warning but it was before adding code in the block so it must be without the underscore as you say.

ncitron:
Also can you please resolve the CI erros?

glihm:
> Also can you please resolve the CI errors?

Yep, I am on it. Thank you for the comment, I'll push once tested. 👍

glihm:
Just pushed the clippy fix, sorry for that. Next time I'll previously run clippy correctly.

ncitron (approved):
LGTM!

Issue #226, opened 2023-04-13
regression for wasm32 target (doesn't compile)


glihm:
@ncitron we have a regression in `wasm32` target due to the [commit 1ae57f1b009efd33365d8fdbb61fde13354fa00b](https://github.com/a16z/helios/commit/1ae57f1b009efd33365d8fdbb61fde13354fa00b).

I don't have the rust knowledge to exactly know how to fix that, because the title of the commit
includes "fix". So does this mean the precedent code wasn't working too?

The idea is that, you are excluding `tokio` for `wasm32` target in all configuration files.
So adding a macro to only include this part of the code in targets that are not `wasm32` looks like a solution.

But, what should we implement in the case it's `wasm32`?

You changed

```rust
// previous

let handle = thread::spawn(move || {
    let account_fut = execution.get_account(&address, Some(&slots), &payload);
    block_on(account_fut)
});

handle.join().unwrap()
```

to this:

```rust
// current

let handle = spawn_blocking(move || {
    block_on(execution.get_account(&address, Some(&slots), &payload))
});

block_on(handle)?
```

If the `current` is not valid for `wasm32`, should we insert the `previous` snippet when it's `wasm32` and the `current` in all other cases?

```rust

#[cfg(not(target_arch = "wasm32"))]
fn get_account(&mut self, address: Address, slots: &[H256]) -> Result<Account> {
    let execution = self.execution.clone();
    let payload = self.current_payload.clone();
    let slots = slots.to_owned();

    let handle = spawn_blocking(move || {
        block_on(execution.get_account(&address, Some(&slots), &payload))
    });

    block_on(handle)?
}

#[cfg(target_arch = "wasm32")]
fn get_account(&mut self, address: Address, slots: &[H256]) -> Result<Account> {
    let execution = self.execution.clone();
    let payload = self.current_payload.clone();
    let slots = slots.to_owned();

    let handle = thread::spawn(move || {
        let account_fut = execution.get_account(&address, Some(&slots), &payload);
        block_on(account_fut)
    });

    handle.join().unwrap()
}
```

Once again, I can't evaluate if this is a solution. Any insights welcome and I can contribute with the PR if needed. 👍

glihm:
Using the proposed fix it compiles fine, for `wasm32` and other targets.
We are the implication of such fix? Are there some side effects I am not aware of?

Thanks a lot!

ncitron:
[...]

PR #227, opened 2023-04-13
fix: use thread instead of tokio when compiled for wasm32

glihm:
As mentioned in the issue #226, it seems to have a regression for wasm32. At least, in beerus when we compile helios as a dependency for wasm32, we had it working fine. But since the commit mentioned in the issue, it's no longer possible due to the fix made with `tokio` in the `evm.rs` from `execution`.

The error we have is `use of undefined crate or module tokio` at `evm.rs:23:5`, which is `use tokio::task::spawn_blocking`.

Here is a proposition, but I don't know the value of it. Hopefully this is something useful, if not, do not hesitate to tell me and I'll close the PR.

What may be an alternative to that PR then?

### DIFF
--- a/execution/src/evm.rs
+++ b/execution/src/evm.rs
@@ -20,7 +20,12 @@ use log::trace;
 use revm::{AccountInfo, Bytecode, Database, Env, TransactOut, TransactTo, EVM};
 
 use consensus::types::ExecutionPayload;
+
+#[cfg(not(target_arch = "wasm32"))]
 use tokio::task::spawn_blocking;
+
+#[cfg(target_arch = "wasm32")]
+use std::thread;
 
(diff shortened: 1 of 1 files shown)


glihm:
Maybe [futures executor](https://docs.rs/futures/0.3.1/futures/executor/fn.block_on.html) may be suitable. I'll try to check that in our implementation.

glihm:
I am trying to setup an helios env to test easily `wasm32` build. But the changes I made use directly the executor, instead of spawning the thread.

glihm:
Thank you for this I have my setup now and yes same error.

I'll investigate from the issue you linked. I'll get back in the history of related packages and try to find out.

PR #230, opened 2023-04-18
feat: adding support for 32-bit ARM architectures

fmhall:
This PR adds a new build target for the release workflow targeting `armv7l`. This is the standard architecture for all non-64 bit Raspberry Pis, and other microcomputers.

This build was tested against a Raspberry Pi running Raspberry Pi OS (fka Raspbian) Bullseye. 

Note that very old Raspberry Pis may be running Raspbian Buster (the older version of the OS), and targeting them is possible when the version of `GLIBC` the builder VM is running is `2.28` (tested by building locally with `cross-rs` and run on Raspbian Buster). This would require a self-hosted runner for Github Actions, as the oldest version of Ubuntu that GA supports out of the box is 20.02, which is being used here.

If your Raspberry Pi is using Buster (check using `cat /etc/os-release`), consider upgrading to Bullseye with the [Imager](https://www.raspberrypi.com/software/) or in-place like [this](https://www.makeuseof.com/tag/raspberry-pi-update-raspbian-os).

This PR also includes modifications to `heliosup` to properly target `armv7l`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -165,3 +165,51 @@ jobs:
             helios_linux_amd64.tar.gz
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  build-linux-armv7:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: checkout
+        uses: actions/checkout@v1
+
+      - name: install rust nightly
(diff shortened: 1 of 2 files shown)


fmhall:
> LGTM!

Any ideas on the failed test?

Issue #235, opened 2023-04-28
feat: Testing Documentation


giovannivignone:
The way to test packages is unclear and should be specified in the ReadMe.

PR #236, opened 2023-04-28
feat: Testing Documentation

giovannivignone:
ref: #235 

- Adding testing documentation to the ReadMe.md
- Adding a cargo testall command to run all tests in all workspaces/packages
- Adding a .env.example to make it easy for devs to load envs into their tests
- Adding dotenv to dev dependencies


.env.example
```diff
@@ -0,0 +1,3 @@
+CONSENSUS_RPC=https://www.lightclientdata.org
+MAINNET_RPC=https://eth-mainnet.g.alchemy.com/v2/XXXXX
```
ncitron, line 2: Should we differentiate between rpc types?

Wouldn't it make more sense to have:
MAINNET_CONSENSUS_RPC
MAINNET_EXECUTION_RPC
GOERLI_CONSENSUS_RPC
GOERLI_EXECUTION_RPC

ncitron (commented):
Mostly looks good.

Can you also make sure that if a test uses the env vars, if the env var is not set, just skip it rather than failing?

PRs made from forks cannot access a repos secrets, so the lack of these vars will cause all those external PRs to fail CI, which isn't great. Short term I think skipping them is fine (it will run anyway when it is merged to master though).

.env.example
giovannivignone: Done @ncitron

ncitron (approved):
LGTM!

Issue #232, opened 2023-04-20
feat: support binding rpc to custom address


ncitron:
Right now the rpc binds to `127.0.0.1` which is fine for most uses. However, if you wanted to run Helios on a server and connect to it from another machine, you might want to bind it to another address such as `0.0.0.0` to expose it publicly. We should add a cli and config option to support this.

PR #238, opened 2023-05-04
feat: allow rpc to bind to custom addresses

sagar-a16z:
## Changes 

Added cli support for `--rpc-bind-ip/-b` to set a custom IP for helios to bind to.

fixes https://github.com/a16z/helios/issues/232


ncitron (approved):
LGTM!

PR #239, opened 2023-05-10
Avoid unnecessary clone in NimbusRPC's `get_updates`

whfuyn:
https://github.com/a16z/helios/blob/6bc43bd81b325c3f637714d79401b4420a437839/consensus/src/rpc/nimbus_rpc.rs#L52-L61
Since we have the ownership of the response, we can simply consume it instead of cloning.


ncitron (approved):
LGTM! Nice catch

Issue #258, opened 2023-08-18
Error: block not available: latest


ltfschoen:
If I run the **call** example with `cargo run -p helios --example call` it outputs the following. Note that I have omitted the Alchemy API key from the output with XXXXX
```
   Compiling execution v0.4.1 (/opt/execution)
   Compiling client v0.4.1 (/opt/client)
   Compiling helios v0.4.1 (/opt)
    Finished dev [unoptimized + debuginfo] target(s) in 21.57s
     Running `target/debug/examples/call`
[2023-08-18T00:57:28Z INFO  call] Execution RPC URL: "https://eth-mainnet.g.alchemy.com/v2/XXXXX"
[2023-08-18T00:57:28Z INFO  call] Consensus RPC URL: https://www.lightclientdata.org
[2023-08-18T00:57:28Z INFO  call] ["MAINNET"] Client built with external checkpoint fallbacks
[2023-08-18T00:57:28Z DEBUG reqwest::connect] starting new connection: https://www.lightclientdata.org/
[2023-08-18T00:57:28Z DEBUG call] Context: call @ 0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac::renderBroker(uint256) <5>
[2023-08-18T00:57:28Z DEBUG reqwest::connect] starting new connection: https://eth-mainnet.g.alchemy.com/
[2023-08-18T00:57:29Z DEBUG reqwest::async_impl::client] redirecting 'https://www.lightclientdata.org/eth/v1/beacon/light_client/bootstrap/0x7beab8f82587b1e9f2079beddebde49c2ed5c0da4ce86ea22de6a6b2dc7aa86b' to 'https://beacon-nd-995-871-887.p2pify.com/c9dce41bab3e120f541e4ffb748efa60/eth/v1/beacon/light_client/bootstrap/0x7beab8f82587b1e9f2079beddebde49c2ed5c0da4ce86ea22de6a6b2dc7aa86b'
[2023-08-18T00:57:29Z DEBUG reqwest::connect] starting new connection: https://beacon-nd-995-871-887.p2pify.com/
[2023-08-18T00:57:29Z INFO  call] [ABIGEN] 0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac::renderBroker(uint256) -> Response Length: 106476
[...]

Issue #254, opened 2023-08-16
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: could not fetch bootstrap


ltfschoen:
If this PR https://github.com/a16z/helios/pull/253/files, as at commit f4a7d4c66a770f437c79a0842d9b387544fce5de, if i run the following to build a Docker container using Ubuntu 23.04
```
./docker/docker.sh
```
and then after that script enters me into the container i run 
```
cargo run -p helios --example client
```
it outputs the following:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.65s
     Running `target/debug/examples/client`
Constructed client!
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: could not fetch bootstrap

Location:
    /opt/consensus/src/consensus.rs:304:26', /opt/consensus/src/consensus.rs:80:32
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

i'm using version `nightly-2023-07-31` in the Docker container, because the latest master branch uses that version in the rust-toolchain file

ncitron:
In your example, you are using lightclientdata.org, which serves ethereum mainnet data, but you are using the goerli configs and a goerli execution rpc.

Also note that we are working on a couple of PRs to clean a few things up, so the current state of master will be in a bit of a weird spot for the next few days.

ltfschoen:
actually, now that i'm connecting to Goerli with the right endpoint, that error `could not fetch bootstrap` is because it can't get the checkpoint 

i also noticed the instructions in config.md, which say to paste the following in ~/.helios/helios.toml.
so i ran the following
```
mkdir -p ~/.helios
mkdir -p /root/.helios/data/goerli
vim ~/.helios/helios.toml
```

Then pasted the following in ~/.helios/helios.toml
```
[goerli]
# The consensus rpc to use. This should be a trusted rpc endpoint. Defaults to Nimbus testnet.
consensus_rpc = "http://testing.prater.beacon-api.nimbus.team"
# [REQUIRED] The execution rpc to use. This should be a trusted rpc endpoint.
execution_rpc = "https://ethereum-goerli-rpc.allthatnode.com"
# The port to run the JSON-RPC server on. By default, Helios will use port 8545.
rpc_port = 8545
# The latest checkpoint. This should be a trusted checkpoint that is no greater than ~2 weeks old.
[...]

PR #262, opened 2023-08-18
feat: Docker for Helios

ltfschoen:
Allows users on macOS or Linux and possibly other operating systems to run Helios from within a Docker container by following the instructions in the README file.

To use this PR run the following,
```
git clone https://github.com/ltfschoen/helios
cd helios
git fetch origin docker:docker
git checkout docker
```
Then follow the instructions here https://github.com/a16z/helios/pull/262/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R55

Note that the following examples generated errors, but those issues have been raised in issues #254 and #258.

* "Basic" example
```
$ cargo run -p helios --example basic

    Finished dev [unoptimized + debuginfo] target(s) in 0.63s
     Running `target/debug/examples/basic`
[2023-08-18T10:22:25Z INFO  basic] Using untrusted RPC URL [REDACTED]
[2023-08-18T10:22:25Z INFO  basic] Using consensus RPC URL: https://www.lightclientdata.org
[2023-08-18T10:22:25Z INFO  basic] Built client on network "MAINNET" with external checkpoint fallbacks
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: could not fetch bootstrap rpc error on method: bootstrap, message: error sending request for url (https://www.lightclientdata.org/eth/v1/beacon/light_client/bootstrap/0x69e3bb0e44b707636ff2c1d3385b20c21013c95f867c296c7ee84960edcbd103): error trying to connect: dns error: task 9 was cancelled

Location:
    /root/helios/consensus/src/consensus.rs:304:26', /root/helios/consensus/src/consensus.rs:80:32
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: block not available: latest

Location:
    client/src/node.rs:217:23
```

* "Call" example output
```
$ cargo run -p helios --example call

    Finished dev [unoptimized + debuginfo] target(s) in 0.64s
     Running `target/debug/examples/call`
[2023-08-18T10:23:02Z INFO  call] Execution RPC URL: "https://eth-mainnet.g.alchemy.com/v2/XXXXX"
[2023-08-18T10:23:02Z INFO  call] Consensus RPC URL: https://www.lightclientdata.org
[2023-08-18T10:23:02Z INFO  call] ["MAINNET"] Client built with external checkpoint fallbacks
[2023-08-18T10:23:02Z DEBUG reqwest::connect] starting new connection: https://www.lightclientdata.org/
[2023-08-18T10:23:02Z DEBUG call] Context: call @ 0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac::renderBroker(uint256) <5>
[2023-08-18T10:23:02Z DEBUG reqwest::connect] starting new connection: https://eth-mainnet.g.alchemy.com/
[2023-08-18T10:23:03Z INFO  call] [ABIGEN] 0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac::renderBroker(uint256) -> Response Length: 106476
[2023-08-18T10:23:04Z INFO  call] [ABIGEN] 0x8bb9a8baeec177ae55ac410c429cbbbbb9198cac::renderBroker(uint256, uint256) -> Response Length: 64549
[2023-08-18T10:23:04Z DEBUG call] Calling helios client on block: Latest
Error: block not available: latest

Location:
    /rustc/a17c7968b727d8413801961fc4e89869b6ab00d3/library/core/src/convert/mod.rs:716:9
```

* "Client" example output
```
$ cargo run -p helios --example client

    Finished dev [unoptimized + debuginfo] target(s) in 0.78s
     Running `target/debug/examples/client`
Constructed client!
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: could not fetch bootstrap rpc error on method: bootstrap, message: error sending request for url (https://www.lightclientdata.org/eth/v1/beacon/light_client/bootstrap/0x0ed5b225c09f70c1adcdd96f6355f6ca7bf8b8dc7ab7deda9acad97c4cf6224f): error trying to connect: dns error: task 8 was cancelled

Location:
    /root/helios/consensus/src/consensus.rs:304:26', /root/helios/consensus/src/consensus.rs:80:32
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


docker/docker.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+trap "echo; exit" INT
+trap "echo; exit" HUP
+
+PARENT_DIR=$( echo $(dirname "$(dirname "$(realpath "${BASH_SOURCE[0]}")")") )
```
Elvis339, line 6: Hey Luke, nice PR thank you for your effort.

`realpath` by default is not installed on MacOS, I had to run `brew install coreutils`, this might be helpful for other people.

An alternative approach could be a check to see if the OS is macOS (Darwin as returned by uname), it will use the Python code to resolve the path.

```bash
function realpath() {
    if [[ "$(uname)" == "Darwin" ]]; then
        python -c "import os; print(os.path.realpath('$1'))"
    else
        echo "$1"
    fi
}
```

ltfschoen: thanks, that's very interesting feedback and an interesting solution. i initially used that approach here https://github.com/paritytech/substrate/commit/7c3d27927577b7fa9ea659aed3197d1f67dd663e since their previous approach wasn't handling symlinks, and i've since used that approach elsewhere.

it seems `os.path.realpath` also handles symlinks.

i also think most operating systems would already have python installed, and your script uses very basic python syntax so it would be fine with those with python 2.7 and recent versions, but for example i'm on a new macOS machine and i didn't have the `python` binary, but i had `python3` instead.

another possible option is to do it like what is done here https://getsubstrate.io/, and installing as shown here https://docs.brew.sh/Installation, 
```
if [[ "$OSTYPE" == "darwin"* ]]; then
	set -e
	echo "Mac OS (Darwin) detected."

	if ! which brew >/dev/null 2>&1; then
                 echo "Installing Homebrew..."
		/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
	fi

	brew update
	brew install coreutils
fi
```
but they might be prompted to enter a password if they're not an admin
```
Checking for `sudo` access (which may request your password)...
Password:
```

Elvis339: Yeah, installing `coreutils` is also a possibility, it is my preference to maybe ask before installing it because `coreutils` is installing multiple binaries but, up to you.

ltfschoen: i've pushed this commit to prompt macOS users to press SPACE if they want to proceed with the script and install the necessary dependencies https://github.com/a16z/helios/pull/262/commits/445f2cee2ef35ded3ae26803ea02af63e21a946f

docker/docker.sh
```diff
@@ -0,0 +1,79 @@
...
+    read -n1 -r
+    [[ $REPLY == ' ' ]] && break
+  done
+  printf "\n*** Continuing..."
+
+  if ! which brew >/dev/null 2>&1; then
+    printf "\n*** Installing Homebrew..."
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+  fi
+
```
Elvis339, line 23:
```bash
  if ! which realpath >/dev/null 2>&1; then
    brew update
    printf "\n*** Installing coreutils..."
    brew install coreutils
  fi
```

Run `brew update` and `brew install coreutils` if `realpath` is not installed.

ltfschoen: i pushed another commit here for review https://github.com/a16z/helios/pull/262/commits/5554620dd066756654b8081783a30ad6dafa5de6

ncitron:
Thanks for the effort here @ltfschoen! While I appreciate the work, I'm not sure we want to support a docker build system for helios. We expect users to run it on their machines directly, especially since helios will hopefully be used by less technical users who don't have docker on their machines, which is why we have the heliosup installer to make things as simple as possible.

I did however really like the extra detail on how to find checkpoints that you added to the readme, and would appreciate it if you opened a PR to add that elsewhere in the readme.

jgresham:
I could add Helios to NiceNode quickly if there was a published container!

PR #264, opened 2023-08-21
feat: use stable rust toolchain

zvolin:
Remove the 'different-binary-name' feature.
Remove the 'more_qualified_paths' feature.
Unexport the 'superstruct_ssr' macro and keep it private.


zvolin:
I have checked the ba08cc1a to see if there were any special reasons to use the `different-binary-name` but found nothing unclear. Names which are valid rust identifiers are supported on stable so this change should be transparent.
With `different-binary-name` cargo produces also the `cli` binary in `target` but I haven't found it used anywhere.

Regarding the `more_quilified_paths', I don't think this macro was needed in the public API of the `consensus` so I just removed the export and relaxed the token matchers so that it doesn't need this feature.

ncitron:
Thanks for looking into this!

I really appreciate the macro help, as I was struggling to get that working without the unsafe feature.

For the binary name feature, the reason we use it is to ensure that when building the `cli` crate, the output executable is named helios instead of cli. I'm not opposed to this being the case, but it simplified our ci, since when we do produce builds to cut new releases, the binary should be named helios. If we can modify the ci to rename the binary, or some to some other solution to this, I think this would be a great addition.

zvolin:
I've rebased and fixed conflicts.

> For the binary name feature, the reason we use it is to ensure that when building the cli crate, the output executable is named helios instead of cli

yeah I'm aware of that. You can freely rename binaries without nightly features as long as the name is a valid rust identifier, the nightly feature is mostly for cases when it's not as far as I understood. Anyway, this change persists the `helios` binary name, feel free to check it yourself by building.
https://github.com/a16z/helios/pull/264/files#diff-6057d83fb675b27e99f187fa354690b6597018c390cbd50eac854a56233d2addR7

zvolin:
Hey @ncitron, I believe that having helios working on the stable is a nice value, especially when used as a library. Whenever you'll have a moment, I'd love to get your thoughts on this. I believe the change should be transparent except removing `superstruct_ssz` from the public API. Thanks in advance

ncitron (approved):
LGTM! Thanks for the PR!

PR #266, opened 2023-08-23
feat(rpc): Add filter rpc endpoints

PatStiles:
Contributes to #200 . Adds

- `eth_getFilterChanges`
- `eth_uninstallFilter`
- `eth_getNewFilter`
- `eth_getNewBlockFilter`
- `eth_getNewPendingTransactionFilter`

@ncitron wanted your thoughts on adding tests. Some of these are just wrappers around ethers::providers.

### DIFF
--- a/execution/src/execution.rs
+++ b/execution/src/execution.rs
@@ -241,10 +241,7 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
         Ok(logs)
     }
 
-       pub async fn get_filter_changes(
-        &self, 
-        filter_id: &U256,
-    ) -> Result<Vec<Log>> {
+    pub async fn get_filter_changes(&self, filter_id: &U256) -> Result<Vec<Log>> {
         let logs = self.rpc.get_filter_changes(filter_id).await?;
         if logs.len() > MAX_SUPPORTED_LOGS_NUMBER {
             return Err(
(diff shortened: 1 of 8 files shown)


execution/src/execution.rs
```diff
@@ -269,6 +269,83 @@ impl<R: ExecutionRpc> ExecutionClient<R> {
...
+        filter_id: &U256,
+    ) -> Result<Vec<Log>> {
+        let logs = self.rpc.get_filter_changes(&filter_id).await?;
+        if logs.len() > MAX_SUPPORTED_LOGS_NUMBER {
+            return Err(
+                ExecutionError::TooManyLogsToProve(logs.len(), MAX_SUPPORTED_LOGS_NUMBER).into(),
+            );
+        }
+
+        for (_pos, log) in logs.iter().enumerate() {
```
ncitron, line 284: Can we dedupe this from the `get_logs` function? Maybe we can extract the duplicated code in a `verify_logs` function that takes the list of unverified logs as an input?

ncitron (commented):
This looks great!

I think since these are all built on top of existing functionality we can omit new tests for the time being.

execution/src/execution.rs
PatStiles: Done!

PR #273, opened 2023-09-11
fix: bench marks

samtvlabs:
This PR fixes the benchmarks , the the structure of the code had changed , and it seems like it wasnt updated


ncitron (approved):
LGTM!

Issue #272, opened 2023-09-02
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: rpc error on method: finality_update


ltfschoen:
i was running a node with 
```
cargo run -- \
  --network goerli \
  --consensus-rpc http://testing.prater.beacon-api.nimbus.team \
  --execution-rpc https://ethereum-goerli-rpc.allthatnode.com \
  --checkpoint 0x3754cde3eabb1815c15d78ef27a279d4abbd3ece0f720535acd5ca80b3a16f91
```

and then got this error:
```
[2023-09-02T10:58:02Z INFO  consensus::consensus] updated head               slot=6428689  confidence=78.12%  age=00:00:00:14
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: rpc error on method: finality_update, message: error sending request for url (http://testing.prater.beacon-api.nimbus.team/eth/v1/beacon/light_client/finality_update): connection closed before message completed

Location:
    consensus/src/rpc/nimbus_rpc.rs:66:19', /eip-x/helios/consensus/src/consensus.rs:105:39
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

i simply started the node again and it worked again...

i'm guessing it may have been caused because i was using a public endpoint rather than Alchemy? or is it because i need to configure the light client to connection with the provider to have a different polling interval / retry timeout / network check timeout, or similar?

ncitron:
The request may have failed for many reasons on the rpc end. This does bring up a good point though that we should be using a retry policy for all api requests helios makes so that a simple connection failure will not cause Helios to crash.

I can work on a fix this week.

muang0:
@ncitron Would you like me to look at this as a good first issue? I'm interested in contributing to this repo.

ncitron:
That would be great @muang0!

muang0:
@ncitron any concerns with using the [backoff](https://crates.io/crates/backoff) crate? Took an initial [pass](https://github.com/a16z/helios/compare/master...muang0:helios:master#diff-56f8bb55d7972a5310efa5cc9b974ca2b3c4ac8b38ceca8ef04a64b6b3c8fbf1) at it.

ncitron:
That crate looks fine.

ncitron:
Closed by #274

PR #274, opened 2023-09-11
Add retry logic for api calls

muang0:
Retry logic added for consensus & nimbus api calls.

#272

### DIFF
--- a/consensus/src/rpc/nimbus_rpc.rs
+++ b/consensus/src/rpc/nimbus_rpc.rs
@@ -15,6 +15,18 @@ pub struct NimbusRpc {
     rpc: String,
 }
 
+async fn get(req: &str) -> Result<reqwest::Response, reqwest::Error> {
+    retry_notify(
+        ExponentialBackoff::default(),
+        || async {
+            Ok(reqwest::get(req).await?)
+        },
+        |e, dur| {
+            warn!("rpc error occurred at {:?}: {}", dur, e)
+        }
+    ).await
+}
+
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 impl ConsensusRpc for NimbusRpc {
@@ -33,7 +45,7 @@ impl ConsensusRpc for NimbusRpc {
 
         let client = reqwest::Client::new();
         let res = client
-            .get(req)
+            .get(req)   // TODO implement backoff
             .send()
             .await
             .map_err(|e| RpcError::new("bootstrap", e))?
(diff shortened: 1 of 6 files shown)


muang0:
Tested by building & running locally. Happy to look at implementing a more comprehensive testing strategy for these changes if we want.

ncitron (changes requested):
This is looking great!

Can we just do a little deduplicating on some of this code?

I think we can probably have one helper function in `NimbusRpc` with a signature like `async fn get<R: Serialize>(req: &str) -> Result<R>`.

In `CheckpointFallback` we can probably do a similar thing with `async fn get_checkpoints(constructed_url: &str) -> Result<reqwest::Response>`.

muang0:
> This is looking great!
> 
> Can we just do a little deduplicating on some of this code?
> 
> I think we can probably have one helper function in `NimbusRpc` with a signature like `async fn get<R: Serialize>(req: &str) -> Result<R>`.
> 
> In `CheckpointFallback` we can probably do a similar thing with `async fn get_checkpoints(constructed_url: &str) -> Result<reqwest::Response>`.

I'm following you regarding the CheckpointFallback function signature, but I'm struggling to understand the NimbusRpc signature. I'm not sure how we would get the generic funtion to deserialize into the specific type we're looking for ie. `OptimisticUpdateResponse`, `BeaconBlockResponse`, etc. Do we want to handle deserialization of the response outside of the abstracted `get` function similar to in `CheckpointFallback`? Thanks for the help.

ncitron:
I think this is in the right direction, but we should be able to do a bit better.

I think the `.json<SomeOutputType>` part can be modified to `.json<R>` in this example, assuming that `R` is a generic on the `get` function that implements `DeserializeOwned` (judging from the signature of the [json method of reqwest::Response](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.json)).

muang0:
> I think this is in the right direction, but we should be able to do a bit better.
> 
> I think the `.json<SomeOutputType>` part can be modified to `.json<R>` in this example, assuming that `R` is a generic on the `get` function that implements `DeserializeOwned` (judging from the signature of the [json method of reqwest::Response](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.json)).

Doh, of course! I was getting hung up on how to reference the generic from within the function. Think I've got it in a ready-to-review state now.

consensus/src/rpc/nimbus_rpc.rs
```diff
@@ -28,15 +43,7 @@ impl ConsensusRpc for NimbusRpc {
...
-        let client = reqwest::Client::new();
-        let res = client
-            .get(req)
-            .send()
-            .await
-            .map_err(|e| RpcError::new("bootstrap", e))?
-            .json::<BootstrapResponse>()
-            .await
-            .map_err(|e| RpcError::new("bootstrap", e))?;
+        let res: BootstrapResponse = get(&req).await?;
```
ncitron, line 46: Looks great. Can we just make sure that the error types here are all `RpcError` like they were originally. It makes it easier when something goes wrong to track from exactly which rpc call it came from.

ncitron:
Oh also can you run a quick `cargo fmt` so that CI passes?

PR #275, opened 2023-09-15
feat(tracing): Switch from logs to tracing

### LINKED ISSUE #181: consider switching from `log` crate to `tracing`
Given that internal lib usage seems to be a goal of the project, `log` is an anti-feature. [tracing](https://docs.rs/tracing/latest/tracing/) should be used instead, to give consuming binaries better control over events. Tracing offers drop-in replacement macros for `log`, so integration should be smooth

PatStiles:
Closes #181 Adds drop in tracing. Leaves adding tracing subscribers for future discussion.


ncitron:
LGTM! Can you just run `cargo fmt` so that we can pass CI?

PatStiles:
Done!

ncitron (approved):
LGTM

PR #283, opened 2023-10-17
Update LICENSE

davidcardenasus:


ncitron:
My understanding is that putting just the year of creation in the copyright is standard.

PR #290, opened 2023-10-24
Replace `wasm-timer` with `zduny-wasm-timer`

orenyomtov:
This fixes the following error that happens when running `helios-ts` in a Web Worker environment:

telegram-cloud-photo-size-1-5023824917329718351-y

More information on the origin of the error:
https://github.com/tomaka/wasm-timer/issues/12

And this is the commit of zduny's fork that fixed it:
https://github.com/zduny/wasm-timer/commit/31beb5c336a52e0438594fcfea45c164fb450350

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -686,7 +686,7 @@ dependencies = [
 
 [[package]]
 name = "cli"
-version = "0.4.1"
+version = "0.5.0"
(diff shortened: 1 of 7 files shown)


ncitron (approved):
LGTM!

PR #300, opened 2023-10-29
Exposing rpc_bind_ip on ClientBuilder

rkreutz:
Hello,

while updating [HeliosKit](https://github.com/rkreutz/HeliosKit) I've noticed that `rpc_bind_ip` is not accessible from `ClientBuilder`, only way to inject it into the builder would be to serialise the config with the binding IP which might be cumbersome, exposing it directly through the builder makes it easier to override.


ncitron (approved):
Good catch!

PR #308, opened 2023-12-08
fix(example/call): read local env variables via dotenv

hsyodyssey:
add `dotenv().ok();` in order to read env variables from .env located in the current project folder.

### DIFF
--- a/examples/call.rs
+++ b/examples/call.rs
@@ -1,11 +1,11 @@
 #![allow(deprecated)]
 
+use dotenv::dotenv;
 use ethers::prelude::*;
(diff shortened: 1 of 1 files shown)


ncitron (approved):
Good catch! LGTM!

ncitron:
Looks like CI isn't passing. Can you make sure to run `cargo fmt` so our formatter checks pass?

hsyodyssey:
> Looks like CI isn't passing. Can you make sure to run `cargo fmt` so our formatter checks pass?

updated ser.

Issue #293, opened 2023-10-24
feat: add database that uses browser local storage for caching checkpoints


ncitron:
Right now the only WASM supported checkpoint database is `ConfigDB` which just returns the hardcoded checkpoint from the config. It would be useful to have a database that cache checkpoints in local storage.

PR #312, opened 2023-12-26
feat: add local storage cache

qiweiii:
resolve: #293 

- added a db that use local storage in helios-ts

Not sure how to test this, I tried to run the `run.sh` but there is network error, it could be my network problem since I am using a vpn

image

### DIFF
--- a/helios-ts/lib.ts
+++ b/helios-ts/lib.ts
@@ -31,7 +31,7 @@ export class HeliosProvider {
   async waitSynced() {
     await this.#client.wait_synced();
   }
-  
+
   async request(req: Request): Promise<any> {
     switch(req.method) {
       case "eth_getBalance": {
--- a/helios-ts/lib.ts
+++ b/helios-ts/lib.ts
@@ -1,7 +1,8 @@
 import init, { Client } from "./pkg/index";
 
-
-export async function createHeliosProvider(config: Config): Promise<HeliosProvider> {
+export async function createHeliosProvider(
+  config: Config
+): Promise<HeliosProvider> {
   const wasmData = require("./pkg/index_bg.wasm");
   await init(wasmData);
   return new HeliosProvider(config);
@@ -19,8 +20,15 @@ export class HeliosProvider {
     const consensusRpc = config.consensusRpc;
     const checkpoint = config.checkpoint;
     const network = config.network ?? Network.MAINNET;
+    const dbType = config.dbType ?? "local_storage";
 
-    this.#client = new Client(executionRpc, consensusRpc, network, checkpoint);
+    this.#client = new Client(
+      executionRpc,
+      consensusRpc,
+      network,
+      checkpoint,
+      dbType
+    );
     this.#chainId = this.#client.chain_id();
   }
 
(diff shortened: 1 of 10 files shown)


ncitron:
Hmm yeah that is a strange error. Will try testing myself to see if it works on my machine.

helios-ts/src/storage.rs
```diff
@@ -0,0 +1,45 @@
...
+use config::Config;
+use consensus::database::Database;
+use eyre::Result;
+
+#[derive(Clone)]
+pub struct StorageDB {
+    checkpoint: Vec<u8>,
+}
+
+impl Database for StorageDB {
```
ncitron, line 13: Can we rename this to `LocalStorageDB` to make it clear that it is the browser local storage.

helios-ts/src/storage.rs
```diff
@@ -0,0 +1,45 @@
...
+                .map(|c| hex::decode(c).unwrap())
+                .unwrap_or(config.default_checkpoint.clone());
+            return Ok(Self { checkpoint });
+        }
+
+        eyre::bail!("local_storage not available")
+    }
+
+    fn load_checkpoint(&self) -> Result<Vec<u8>> {
+        Ok(self.checkpoint.clone())
```
ncitron, line 31: Lets move the fetching logic in the constructor into here so we can ensure we have the freshest checkpoint. It also saves us from needing to store the checkpoint and have an unnecessary clone.

helios-ts/src/storage.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    fn save_checkpoint(&self, checkpoint: &[u8]) -> Result<()> {
+        let window = web_sys::window().unwrap();
+        if let Ok(Some(local_storage)) = window.local_storage() {
+            local_storage
+                .set_item("checkpoint", &hex::encode(checkpoint))
+                .unwrap();
```
ncitron, line 39: Can we return a err result here?

helios-ts/src/lib.rs
```diff
@@ -21,7 +24,7 @@ macro_rules! log {
 
 #[wasm_bindgen]
 pub struct Client {
```
ncitron, line 26: Lets come up with a way to allow the typescript to set what type of db we want to use here. Maybe we can do something with generics if we can get js and ts to interop with them well?

helios-ts/src/storage.rs
```diff
@@ -0,0 +1,45 @@
...
+pub struct StorageDB {
+    checkpoint: Vec<u8>,
+}
+
+impl Database for StorageDB {
+    fn new(config: &Config) -> Result<Self> {
+        console_error_panic_hook::set_once();
+        let window = web_sys::window().unwrap();
+        if let Ok(Some(local_storage)) = window.local_storage() {
+            let checkpoint = local_storage.get_item("checkpoint").unwrap();
```
ncitron, line 18: Lets return an err result if it fails here.

ncitron (commented):
Looking good!

helios-ts/src/storage.rs
qiweiii: sounds good, does this mean we remove the checkpoint field on LocalStorageDB struct, and in constructor, call save_checkpoint to save to actual local_storage, so it can be load later on?

helios-ts/src/lib.rs
qiweiii: wasm newbie here😅, still learning how to do this. it seems it's easier to just pass 'config' or 'localstorage' to `Client`
```ts
new Client(executionRpc, consensusRpc, network, checkpoint, db || 'localstorage');
```

helios-ts/src/storage.rs
ncitron: I don't think we need to do any saving in the constructor. But yeah we won't need the vec anymore in the struct.

helios-ts/src/lib.rs
qiweiii: If I use generic in `pub struct Client`, it gives an error:

```
structs with #[wasm_bindgen] cannot have lifetime or type parameters currently
```

So I changed to use an enum

helios-ts/src/lib.rs
```diff
@@ -4,24 +4,60 @@ extern crate web_sys;
...
+pub enum DatabaseType {
+    Memory(ConfigDB),
+    LocalStorage(LocalStorageDB),
+}
+
+impl Database for DatabaseType {
+    fn new(config: &Config) -> Result<Self> {
+        // Implement this method based on the behavior of ConfigDB and LocalStorageDB
+        match config.database_type.as_deref() {
+            Some("memory") => Ok(DatabaseType::Memory(ConfigDB::new(config)?)),
```
ncitron, line 37: Can we rename this to `config`?

helios-ts/src/lib.rs
```diff
@@ -4,24 +4,60 @@ extern crate web_sys;
...
+            Some("local_storage") => Ok(DatabaseType::LocalStorage(LocalStorageDB::new(config)?)),
```
ncitron, line 38: And rename this to `localstorage`

helios-ts/lib.ts
```diff
@@ -19,8 +20,15 @@ export class HeliosProvider {
     const consensusRpc = config.consensusRpc;
     const checkpoint = config.checkpoint;
     const network = config.network ?? Network.MAINNET;
+    const dbType = config.dbType ?? "local_storage";
```
ncitron, line 23: Lets make the default `localstorage` since this has the best compatibility across different wasm environments

ncitron (commented):
Just a couple last comments.

Once this is done can you make sure that CI is passing (fmt, clippy, test) and then I think we are good to merge.

ncitron (approved):
LGTM! Awesome work!

PR #326, opened 2024-03-15
Add `SEPOLIA` to `CheckpointFallback` networks

sergey-melnychuk:
Without `SEPOLIA` in the list of networks, the call `CheckpointFallback::get_healthy_fallback_services` fails for `networks::Network::SEPOLIA`.

PR #328, opened 2024-03-15
fix: `test_checkpoint_fallback` test

sergey-melnychuk:
Fix the test after https://github.com/a16z/helios/pull/326


ncitron:
Thanks for following up. Already handles in #327

PR #334, opened 2024-05-23
fix: replace the unavailable goerli testnet part with sepolia and holesky

hsyodyssey:
The closure of the Goerli testnet prevents some related tests and examples from passing.
Replacing them with sepolia and holesky


ncitron (approved):
LGTM!

PR #335, opened 2024-05-23
update dependencies to newest versions

LKozlowski:
This PR updates most of the dependencies to its newest versions.
Only `jsonrpsee` hasn't been updated to the newest version as it requires a bigger refactor, but I bumped it a few versions anyway.


ncitron (approved):
LGTM!

PR #339, opened 2024-07-05
refactor: decouple object methods into reusable primitives

xavierdmello:
Decouples object methods into standalone primitives to enable usage in restricted environments without `async` and `SystemTime`, like ZKVMs. This allows critical functions (e.g., verify_update) to run in synchronous ZK proving environments, facilitating on-chain verification of beacon chain state.


common/src/lib.rs
```diff
@@ -1,3 +1,26 @@
+pub mod consensus;
 pub mod errors;
 pub mod types;
 pub mod utils;
+
+pub mod execution {
```
ncitron, line 6: What is the purpose of moving parts of execution into common?

common/Cargo.toml
```diff
@@ -1,12 +1,29 @@
 [package]
```
ncitron, line 1: It seems like the major goal of this PR is moving much of the core consensus verification functions into it's own crate? Right now that crate is common.

I prefer if we keep common as lightweight as possible and be things that are used by many crates and must be placed in common to avoid the circular import rule. I think the better solution would be to move the core consensus primitives into some new crate `consensus-core` which provers can then consume.

common/Cargo.toml
```diff
@@ -1,12 +1,29 @@
...
 thiserror.workspace = true
+
+ethers-core.workspace = true
+zduny-wasm-timer.workspace = true
+bytes.workspace = true
+
+milagro_bls = { workspace = true, optional = true }
+bls12_381 = { workspace = true, optional = true }
+
+[target.'cfg(target_vendor = "succinct")'.dependencies]
```
ncitron, line 28: Same issue with sp1 specific functionality.

ncitron (commented):
This is a great start! My comments focus on two main areas.

1. Avoid SP1 specific functionality. We want to ensure any prover can use this and would like to stay neutral. Because of this I don't love the SP1 patched crates being directly by Helios.

2. Increasing the size of the common crate. If the major goal here is just to have access to the core consensus functions in a separate crate, I think the best way to do so would be by moving those into some new `consensus-core` crate and deeply assess how to avoid dragging things out of `config` and `execution` into this new crate or common.

consensus/src/consensus.rs
```diff
@@ -41,25 +49,15 @@ pub struct ConsensusClient<R: ConsensusRpc, DB: Database> {
 
 #[derive(Debug)]
 pub struct Inner<R: ConsensusRpc> {
```
ncitron, line 51: Why have all of these fields as well as many methods on this struct been made public when they don't seem to need to be?

ncitron (commented):
This is looking really good. Left a few comments and questions and think we should be good to merge after these things have been addressed.

Also can you address the fmt and clippy issues we are running into in CI?

consensus/src/consensus.rs
xavierdmello: I reverted some fields to be private, but kept `store` and `rpc` public alongside the methods. These values and methods are useful inside other programs when helios is used as a library. 

e.g in sp1-telepathy we pass grab `store` and call `expected_current_slot` from `Inner`

https://github.com/succinctlabs/sp1-telepathy/blob/1fe903bd1507010180f303a2ca3be6ebbd4155ee/script/bin/operator.rs#L184-L189

xavierdmello:
Addressed comments & CI. One notable change was going from .clone() to clone_from() to make clippy happy here: https://github.com/succinctlabs/helios/blob/6e0e4f122038d8b570f9c5af9a834c79ec597c29/consensus-core/src/consensus_core.rs#L149-L155

ncitron (approved):
LGTM! Amazing work!

PR #341, opened 2024-07-12
docs: fix allthatnode ethereum mainnet rpc url

inchori:
The All that Node Ethereum RPC URLs are changed. So I created a pull request.
I have a question. Does the helios keep support `goerli` network in this project? 
Because the All That Node deprecated the goerli testnet.


ncitron (approved):
Thanks for updating this and yes we have deprecated goerli as well but we should probably check for references to it in the docs.

PR #364, opened 2024-09-03
fix: make file_db benchmark runable

hsyodyssey:
**Description**

- Due to the migration from `ether-rs` to `Alloy `and changes in some basic API parameter types, the benchmark part was unable to run properly. 
- **This PR fixes the tests for file_db.** 
- The other benchmark tests involve changes related to the testnet, which would require more extensive modifications. Since the file_db test does not involve network requests, it is submitted as a separate PR

**Local test result**
image


ncitron (approved):
LGTM!

PR #369, opened 2024-09-11
chore: serialize lc store

ratankaliani:
This `Serialize` bound was removed in https://github.com/a16z/helios/pull/356, but is necessary to convert the `LightClientStore` into bytes, which is used in https://github.com/succinctlabs/sp1-helios/.


ncitron (approved):
LGTM!

PR #377, opened 2024-09-17
wip: `no-std` support for `consensus-core` and `common`

ratankaliani:
## Overview
External users of Helios require `common` and `consensus-core` to be compiled to `no-std` targets such as `wasm32-unknown-unknown` for use in chain binaries (e.g. WASM binaries for Substrate chains). Add support for this target to both `consensus-core` and `common` by changing the error crate, adding a `no-std` flag, and adding a CI job which confirms the compatibility of these crates.

## Details
- Add `no-std` flag to both `common` and `consensus-core`.
- Add CI test to confirm that `common` and `consensus-core` compile on the `wasm32-unknown-unknown` target.
- Change error library to `anyhow` which has `no-std` support.
- Use `alloy-primitives` instead of `alloy` for `no-std` support.
- Note: Needed to remove the `ssz-derive` `Decode` trait, because it required some `std` crates. This can be re-added more carefully in the future.
- `eyre::Report` and `anyhow::Result` are equivalent.

### DIFF
--- a/consensus-core/Cargo.toml
+++ b/consensus-core/Cargo.toml
@@ -19,4 +19,6 @@ superstruct.workspace = true
 thiserror-no-std.workspace = true
 tracing.workspace = true
 zduny-wasm-timer.workspace = true
+# Building consensus-core for wasm requires getrandom with the js feature.
+# Source: https://github.com/alloy-rs/core?tab=readme-ov-file#wasm-support
 getrandom = { version = "0.2", features = ["js"] }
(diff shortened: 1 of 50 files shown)


consensus-core/Cargo.toml
```diff
@@ -4,7 +4,7 @@ name = "consensus-core"
 edition = "2021"
 
 [dependencies]
-alloy = { version = "0.2.1", features = ["consensus", "rpc-types", "ssz", "rlp", "k256"] }
+alloy-primitives = { version = "0.7.7", default-features = false }
```
ratankaliani, line 7: Note: `alloy` doesn't have `no-std` support, but `alloy-primitives` does, and this crate only needed `alloy-primitives`.

ratankaliani:
Closed in favor of https://github.com/a16z/helios/pull/378

PR #381, opened 2024-09-30
fixes #380 | fixes get_filter_changes() function

### LINKED ISSUE #380: fix: get_filter_changes() function in client.rs
In the client.rs file, the function there is a function called `get_filter_changes()` which is incorrect. It calls `self.node.uninstall_filter()` instead of `self.node.get_filter_changes()`

{C87940A1-A409-4038-ACE3-0FF668B2144D}

18aaddy:
This PR fixes #380
I have updated the function to return correct value and call the correct function.
Working locally ✅

### DIFF
--- a/core/src/client/mod.rs
+++ b/core/src/client/mod.rs
@@ -121,8 +121,8 @@ impl<N: NetworkSpec, C: Consensus<N::TransactionResponse>> Client<N, C> {
         self.node.get_logs(filter).await
     }
 
-    pub async fn get_filter_changes(&self, filter_id: U256) -> Result<bool> {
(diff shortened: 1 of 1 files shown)


18aaddy:
@ncitron can you review pls?

ncitron:
Sorry we were working on some big changes to Helios so this will need to be redone against the current master. Looks like the bug is still there though. Good catch! The location of that file is now in `core/client/mod.rs`

18aaddy:
Yes sure I'll update the PR

18aaddy:
✅ Rebased this branch

PR #392, opened 2024-10-12
feat: Productionize JavaScript and NPM

roninjin10:
Adds npm publishing and generally updates the javascripts to be more robust

## publish.yml

Runs basic checks like linters before publishing. By default it does a `dry-run` which means it tests that npm publishing will work but doesn't actually publish. To publish you must trigger a publish via a workflow action.

## release.yml

Must have been a bad rebase because this


helios-ts/package.json
```diff
@@ -1,22 +1,62 @@
...
+    "build": "webpack && npm run build:types",
+    "build:types": "npm run build:types:cjs && npm run build:types:esm",
+    "build:types:cjs": "tsc --module CommonJS --moduleResolution node --outDir ./types/cjs && find ./types/cjs -type f -name '*.d.ts' -print0 | xargs -0 -I {} sh -c 'mv \"$1\" \"${1%.d.ts}.d.cts\"' _ {}",
+    "build:types:esm": "tsc --module NodeNext --outDir ./types/esm",
+    "docs": "typedoc",
+    "format": "biome format . --write",
+    "format:check": "biome format .",
+    "lint": "biome check . --write --unsafe",
+    "lint:check": "biome check . --verbose",
+    "lint:package": "bunx publint --strict && attw --pack --ignore-rules cjs-resolves-to-esm",
```
roninjin10, line 40:
```suggestion
    "lint:package": "npx bunx publint --strict && attw --pack --ignore-rules cjs-resolves-to-esm",
```

Copy pasta this should be npx since this repo isn’t using bun

PR #397, opened 2024-10-19
perf: avoid redundant tx receipt fetching in verify_logs

### LINKED ISSUE #396: Redundant tx receipts fetching in verify_logs
https://github.com/a16z/helios/blob/2d795779222f9bd84a49eba067c2829f67f4c865/execution/src/execution.rs#L275

The `verify_logs` method iterates over each log item, fetching the corresponding tx receipt from the RPC and encoding it. There is redundant work being done if multiple logs belong to the same tx.

This can be optimized by first collecting all unique tx hashes and then fetching the receipts for those hashes.

eshaan7:
Closes #396.


ncitron (approved):
LGTM!

PR #398, opened 2024-10-21
feat: handle HTTP errors more gracefully by catching them before deserialization

fmhall:
[ x ] Ensure that other implementations of the ConsensusRPC spec do not send error messages with status code `200`, or handle them
[ x ] Potentially use untagged enums to respond with more structured error responses so that the code and message can be "bubbled up"

### DIFF
--- a/ethereum/src/rpc/http_rpc.rs
+++ b/ethereum/src/rpc/http_rpc.rs
@@ -21,13 +21,21 @@ pub struct HttpRpc {
 }
 
 async fn get<R: DeserializeOwned>(req: &str) -> Result<R> {
-    let bytes = retry(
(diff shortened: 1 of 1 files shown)


fmhall:
Example improvement:

Old error: `ERROR helios::consensus: sync failed err=could not fetch bootstrap: rpc error on method: bootstrap, message: missing field `data` at line 1 column 49`

New error: `ERROR helios::consensus: sync failed err=could not fetch bootstrap: rpc error on method: bootstrap, message: HTTP error 404 Not Found: LC bootstrap unavailable`

ncitron (approved):
LGTM!

PR #413, opened 2024-10-29
fix: save every new checkpoint

### LINKED ISSUE #409: fix: save checkpoints before shutdown
Right now the Ethereum consensus saves a new checkpoint to the DB when the shutdown command is triggered. This is fine in the CLI, but some apps shutdown abruptly like in the browser, so the checkpoint never gets saved. We should instead just save checkpoints whenever we get a more recent one.

0xRampey:
fixes #409


ncitron (approved):
LGTM!

Issue #412, opened 2024-10-28
feat: rpc call `web3_clientVersion`


rapiz1:
Hi thanks for your great work. I'm trying to run an app against the rpc endpoint. I wonder if it can support `web3_clientVersion`?

ncitron:
We don't support that now, but it should be relatively easy to do so.

ncitron:
closed by #414

PR #414, opened 2024-10-29
feat(web3_clientVersion): Adds RPC method to fetch client version

mw2000:
We add the RPC method to fetch the client version via web3_clientVersion. This is also the get_client_version function while calling from a the rust client. Tested with function call in examples/basic.rs.

This should tackle #412

### DIFF
--- a/core/src/client/rpc.rs
+++ b/core/src/client/rpc.rs
@@ -74,8 +74,6 @@ trait EthRpc<TX: TransactionResponse + RpcObject, TXR: RpcObject, R: ReceiptResp
     ) -> Result<U64, ErrorObjectOwned>;
     #[method(name = "getCode")]
     async fn get_code(&self, address: Address, block: BlockTag) -> Result<Bytes, ErrorObjectOwned>;
-    #[method(name = "getClientVersion")]
-    async fn get_client_version(&self) -> Result<String, ErrorObjectOwned>;
     #[method(name = "call")]
     async fn call(&self, tx: TXR, block: BlockTag) -> Result<Bytes, ErrorObjectOwned>;
     #[method(name = "estimateGas")]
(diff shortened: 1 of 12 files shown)


core/src/execution/mod.rs
```diff
@@ -176,6 +176,10 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
             .await
     }
 
+    pub async fn get_client_version(&self) -> Result<String> {
+        self.rpc.get_client_version().await
```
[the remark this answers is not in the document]
mw2000: okay sounds good! I can do that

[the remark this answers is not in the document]
mw2000: Tested this with the suggested changes. Getting the following output

```
...
2024-10-30T02:18:05.282814Z  INFO basic: client version: helios-0.7.0
2024-10-30T02:18:05.282990Z  INFO basic: synced up to block: 21075426
2024-10-30T02:18:05.283237Z  INFO basic: balance of deposit contract: 52486351.455433317711710313
```

core/src/execution/mod.rs
```diff
@@ -176,6 +177,11 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
             .await
     }
 
+    pub async fn get_client_version(&self) -> Result<String> {
+        let helios_version = env::var("CARGO_PKG_VERSION");
```
ncitron, line 181: This method fetches the environment variable at runtime, so it works when you compile Helios yourself, but for a shipped Helios binary it will fail. We should be using `std:env!(VAR_NAME)` which is evaluated at compile time.

mw2000: Done!

core/src/client/rpc.rs
```diff
@@ -74,6 +74,8 @@ trait EthRpc<TX: TransactionResponse + RpcObject, TXR: RpcObject, R: ReceiptResp
     ) -> Result<U64, ErrorObjectOwned>;
     #[method(name = "getCode")]
     async fn get_code(&self, address: Address, block: BlockTag) -> Result<Bytes, ErrorObjectOwned>;
+    #[method(name = "getClientVersion")]
```
ncitron, line 77: Oh sorry one last thing. This should be `web3_clientVersion`, not `eth_getClientVersion`. You can see how we do that with the net vs eth namespaces we already support.

mw2000: Pushed these changes in latest commit

ncitron (approved):
LGTM!

PR #416, opened 2024-10-30
fix(benches): Fixes compiler errors for benches

mw2000:
We fix compiler errors that we see while benchmarking.


mw2000:
A few runtime errors that are still to be fixed, which I wanted some feedback on -

Tests:

I see these tests failing - `consensus::tests::test_verify_finality_invalid_finality` and `consensus::tests::test_verify_update_invalid_finality`. Both fail at `thread 'consensus::tests::test_verify_update_invalid_finality' panicked at ethereum/src/consensus.rs:793:55:` and `thread 'consensus::tests::test_verify_finality_invalid_finality' panicked at ethereum/src/consensus.rs:836:64:` respectively.

They fail because the `verify_finality_update` and `verify_update` calls don't result in errors. How can we make these calls fail so we can pass the tests? 
---

Benchmarks:
Benchmarking also seem to fail at `get_balance`

```
Benchmarking get_balance: Warming up for 3.0000 sthread 'main' panicked at benches/get_balance.rs:35:50:
called `Result::unwrap()` on an `Err` value: out of sync: 1730275786 seconds behind

Location:
    /Users/<user>/Desktop/helios/core/src/client/node.rs:60:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

mw2000:
Believe @0xRampey was also working on the compiler errors within tests. Since he found the issue first, we should merge his PR in. I'll reshape this PR to target the `benches` crate while his PR targets tests

0xRampey:
> They fail because the verify_finality_update and verify_update calls don't result in errors. How can we make these calls fail so we can pass the tests?

Turns out that`LightClientHeader::default()` when passed as finalized header, converts to `None` along the way and slips through `verify_generic_update` without being handled! Hence, the calls are passing.
I've added a new Error to handle that case and replaced  `::default()` header with those from older updates. That seems to be doing the trick.

ncitron (approved):
LGTM!

PR #417, opened 2024-10-30
fix: tests in `helios-ethereum`

### LINKED ISSUE #415: fix: tests failing in `helios-ethereum`
Unit tests for `ConsensusClient` expect more generic arguments

0xRampey:
fixes #415

### DIFF
--- a/ethereum/consensus-core/src/consensus_core.rs
+++ b/ethereum/consensus-core/src/consensus_core.rs
@@ -308,7 +308,7 @@ pub fn verify_generic_update<S: ConsensusSpec>(
             return Err(ConsensusError::InvalidFinalityProof.into());
         }
     } else {
-        return Err(ConsensusError::InvalidFinalizedHeader.into());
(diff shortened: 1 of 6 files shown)


ethereum/consensus-core/src/consensus_core.rs
```diff
@@ -307,6 +307,8 @@ pub fn verify_generic_update<S: ConsensusSpec>(
         } else {
             return Err(ConsensusError::InvalidFinalityProof.into());
         }
+    } else {
+        return Err(ConsensusError::MissingFinalizedHeader.into());
```
ncitron, line 311: It is possible to have an update that does not contain a finalized header (see optimisitc updates).

0xRampey: Ah, right

ncitron:
Can you run `cargo fmt` so that the fmt CI passes?

0xRampey:
Done!

ncitron (approved):
LGTM!

PR #423, opened 2024-11-07
feat: implement eth_getBlockReceipts

eshaan7:
Partially Closes: #200.
Related: https://github.com/a16z/helios/issues/200#issuecomment-2460454087

### DIFF
--- a/core/src/execution/mod.rs
+++ b/core/src/execution/mod.rs
@@ -337,6 +337,7 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
             .collect::<Result<HashSet<_>, _>>()?;
 
         // Collect all (proven) tx receipts as a map of tx hash to receipt
+        // TODO: use get_block_receipts instead to reduce the number of RPC calls?
         let receipts_fut = txs_hash.iter().map(|&tx_hash| async move {
             let receipt = self.get_transaction_receipt(tx_hash).await;
             receipt?.map(|r| (tx_hash, r)).ok_or(eyre::eyre!(
--- a/core/src/execution/mod.rs
+++ b/core/src/execution/mod.rs
@@ -243,7 +243,7 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
         let expected_receipt_root = B256::from_slice(&expected_receipt_root.to_fixed_bytes());
 
         if expected_receipt_root != block.receipts_root {
-            return Err(ExecutionError::BlockReceiptRootMismatch(tag).into());
+            return Err(ExecutionError::BlockReceiptsRootMismatch(tag).into());
         }
 
         Ok(Some(receipts))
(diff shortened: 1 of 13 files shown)


core/src/execution/mod.rs
```diff
@@ -197,6 +197,7 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
 
         let tx_hashes = block.transactions.hashes();
 
+        // TODO: replace this with rpc.get_block_receipts?
```
eshaan7, line 200: @ncitron; thoughts on this?

Using `getBlockReceipts` instead of `getTransactionReceipt` would significantly reduce the number of HTTP calls made to the RPC.

eshaan7, line 200: @ncitron; thoughts on this?

Using `getBlockReceipts` instead of `getTransactionReceipt` would significantly reduce the number of HTTP calls made to the RPC.

eshaan7:
The `rpc_equivalence` tests are passing for me locally but failing in GitHub actions. Not sure why.

image

ncitron:
Don't worry about those tests. They are failing in CI because CI runs triggered from forks don't have access to gh secrets (where we keep our alchemy api key) for security reasons.

core/src/execution/mod.rs
eshaan7: https://github.com/ethereum/execution-apis/pull/438

erigon, [geth](https://github.com/ethereum/go-ethereum/pull/27702), [besu](https://github.com/hyperledger/besu/pull/5771), [nimbus](https://github.com/status-im/nimbus-eth1/issues/1945), [reth](https://github.com/paradigmxyz/reth/pull/3321), 

[infura](https://docs.infura.io/api/networks/blast/json-rpc-methods/eth_getblockreceipts), [quicknode](https://www.quicknode.com/docs/ethereum/eth_getBlockReceipts), [alchemy](https://docs.alchemy.com/reference/eth-getblockreceipts), [chainstack](https://docs.chainstack.com/reference/ethereum-getblockreceipts)

yep, seems to be widely supported.

ncitron: Yeah in that case lets do it. Thanks for finding this optimization!

core/src/execution/mod.rs
```diff
@@ -337,6 +337,7 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
             .collect::<Result<HashSet<_>, _>>()?;
 
         // Collect all (proven) tx receipts as a map of tx hash to receipt
+        // TODO: use get_block_receipts instead to reduce the number of RPC calls?
```
eshaan7, line 340: Can significantly optimize `verify_logs` too. Will create a separate issue for it once this PR is merged.

ncitron (approved):
LGTM! Lets do the optimization on getLogs too when you get a chance.

core/src/execution/mod.rs
eshaan7: Issue: https://github.com/a16z/helios/issues/438

PR #424, opened 2024-11-08
Fetch and verify unsafe signer address

### LINKED ISSUE #408: feat: support pulling opstack system config for ethereum light client
Right now Helios's OP Stack only uses hard codes sequencer unsafe signer keys. This is fine, but they can in theory change, so it would be nice to be able to fetch them from the L1 light client using the OP Stack's `SystemConfig` contract.

I'd like to avoid having a requirement of the full Ethereum execution RPC, so think it would be best just to use Ethereum's consensus backend and throw together a special endpoint on the opstack consensus server for providing the single merkle proof we need and verifying it against the most recent header fetched from the Ethereum consensus.

0xRampey:
fixes #408


opstack/src/consensus.rs
```diff
@@ -136,6 +152,79 @@ impl Inner {
...
+            &account_path,
+            &account_encoded,
+        );
+        if !is_valid {
+            warn!(target: "helios::opstack", "account proof invalid");
+            return Err(eyre!("account proof invalid"));
+        }
+        // with storage proof
+        let storage_proof = proof.storage_proof[0].clone();
+        let key = storage_proof.key.0;
```
ncitron, line 200: We need to ensure that the key is actually the unsafe signer location. As implemented now the server can return any slot in the contract and it will be interpreted as the signer.

opstack/src/consensus.rs
```diff
@@ -33,13 +41,15 @@ impl ConsensusClient {
...
             server_url: config.consensus_rpc.to_string(),
-            unsafe_signer: config.chain.unsafe_signer,
+            unsafe_signer: Arc::new(Mutex::new(config.chain.unsafe_signer)),
             chain_id: config.chain.chain_id,
             latest_block: None,
             block_send,
             finalized_block_send,
         };
 
+        verify_unsafe_signer(inner.server_url.clone(), inner.unsafe_signer.clone());
```
ncitron, line 51: Not awaiting this future means it won't necessarily run. We should use the `run` function  to ensure it happens in the background.

opstack/src/server/mod.rs
```diff
@@ -1,8 +1,14 @@
...
-use axum::{extract::State, routing::get, Json, Router};
+use alloy::{primitives::Address, rpc::types::EIP1186AccountProofResponse};
+use axum::{
+    extract::{Path, State},
+    http::StatusCode,
+    routing::get,
+    Json, Router,
+};
 use eyre::Result;
+use revm::primitives::B256;
```
ncitron, line 11: nit: lets use the alloy variant of this for clarity (should be same as it is just a reexport)

opstack/src/server/mod.rs
```diff
@@ -44,6 +61,7 @@ pub async fn start_server(
     let router = Router::new()
         .route("/latest", get(latest_handler))
         .route("/chain_id", get(chain_id_handler))
+        .route("/signer_proof/:block_number", get(signer_proof_handler))
```
ncitron, line 64: I think it would be safer if we pinned the proof to the block hash rather than the number. It's possible that the block at number n is different for the server and the light client (do to a fork).

opstack/src/consensus.rs
```diff
@@ -136,6 +152,79 @@ impl Inner {
...
 
+fn verify_unsafe_signer(server_url: String, signer: Arc<Mutex<Address>>) {
+    #[cfg(not(target_arch = "wasm32"))]
+    let run = tokio::spawn;
+
+    #[cfg(target_arch = "wasm32")]
+    let run = wasm_bindgen_futures::spawn_local;
+
+    run(async move {
+        let eth_config = mainnet();
```
ncitron, line 163: We should at a minimum support setting a different checkpoint (the `-w` flag) and fetching the fallback (the `-l` flag) in the cli. Lets add those and make sure to document what it means (that it is for the ethereum sync) in the comments above the flag param in the cli struct (clap uses this to generate `--help`.

opstack/src/consensus.rs
```diff
@@ -136,6 +152,79 @@ impl Inner {
...
+            Arc::new(eth_config.into()),
+        )?;
+        let block = eth_consensus
+            .block_recv()
+            .unwrap()
+            .recv()
+            .await
+            .ok_or_eyre("failed to receive block")?;
+        // Query proof from op consensus server
+        let req = format!("{}signer_proof/{}", server_url, block.number.to::<u64>());
```
ncitron, line 178: nit: `unsafe_signer_proof`

ncitron (changes requested):
Good work!

Added some comments for some minor changes.

opstack/src/consensus.rs
0xRampey: Ah yes. Changing that

opstack/src/consensus.rs
0xRampey: We use the run func inside it here https://github.com/a16z/helios/blob/7677f593b7d396c7b2e9b90b9dba58afedba6ef8/opstack/src/consensus.rs#L162

opstack/src/server/mod.rs
0xRampey: Makes sense

opstack/src/builder.rs
```diff
@@ -82,7 +82,10 @@ impl OpStackClientBuilder {
                 chain: ChainConfig {
```
ncitron, line 82: We should support building based on network similar to how we do it in Ethereum (see the ethereum builder for inspiration). So we should have a `network` method on the builder and pass in a `Network` enum to properly set everything from within the `ChainConfig`.

cli/src/main.rs
```diff
@@ -191,6 +191,20 @@ struct OpStackArgs {
     execution_rpc: Option<String>,
     #[clap(short, long, env)]
     consensus_rpc: Option<String>,
+    #[clap(
+        short = 'w',
+        long = "eth-checkpoint",
```
ncitron, line 196: nit: ethereum-checkpoint

cli/src/main.rs
```diff
@@ -191,6 +191,20 @@ struct OpStackArgs {
...
+        env = "ETH_CHECKPOINT",
```
ncitron, line 197: `ETHEREUM_CHECKPOINT`

cli/src/main.rs
```diff
@@ -191,6 +191,20 @@ struct OpStackArgs {
...
+        help = "Set custom weak subjectivity checkpoint for Ethereum mainnet. Helios uses this to sync and trustlessly fetch the correct unsafe signer address used by <NETWORK>"
+    )]
+    checkpoint: Option<B256>,
+    #[clap(
+        short = 'l',
+        long = "eth-load-external-fallback",
```
ncitron, line 203: `ethereum-load-external-fallback`

cli/src/main.rs
```diff
@@ -191,6 +191,20 @@ struct OpStackArgs {
...
+        env = "ETH_LOAD_EXTERNAL_FALLBACK",
```
ncitron, line 204: ETHEREUM_LOAD_EXTERNAL_FALLBACK

ncitron (changes requested):
Just a few more small things and then we're good to go. Also sorry for the delay. Got behind on reviews during devcon.

opstack/src/server/mod.rs
```diff
@@ -62,18 +85,44 @@ async fn chain_id_handler(State(state): State<Arc<RwLock<ServerState>>>) -> Json
...
+        B256::from_str(UNSAFE_SIGNER_SLOT).map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+    let block = rpc
+        .get_block(block_hash)
+        .await
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+    let proof = rpc
+        .get_proof(
+            state.read().await.system_config_contract,
+            &[signer_slot],
+            block.number.to(),
```
ncitron, line 104: Also I think its possible to pass a blockhash in here. We should prefer that in case there is a fork and multiple blocks exist at a given block number.

0xRampey: Makes sense!

0xRampey:
> Also sorry for the delay. Got behind on reviews during devcon.

Figured. Sweet talk at Devcon btw!

Lmk if there are any more nits

opstack/src/builder.rs
```diff
@@ -18,6 +19,7 @@ pub struct OpStackClientBuilder {
     consensus_rpc: Option<Url>,
     execution_rpc: Option<Url>,
     rpc_socket: Option<SocketAddr>,
+    eth_network: Option<EthNetwork>,
```
ncitron, line 22: Oh here I had meant we should support passing in the opstack `Network` which we can use to fill out the `ChainConfig` values that the builder needs, but good catch that we also need to know the Ethereum network that it is built on. Maybe we can add `eth_network` as a parameter to the opstack `ChainConfig` and then just allow setting the opstack `Network` with a `network` method on the builder, which is then used to grab the correct `ChainConfig` so that we don't need to set things like the chain id, system config address, and default unsafe signer manually.

ncitron (changes requested):
Ok one last thing then we're good to go.

opstack/src/builder.rs
0xRampey: Yeah that's cleaner

ncitron (approved):
LGTM!

PR #425, opened 2024-11-11
fixes #233: eth_getTransactionByHash & eth_getTransactionByBlockHashAndIndex in helios-ts provider

eshaan7:
Closes #233.

Changes:
- Expose `eth_getTransactionByHash` and `eth_getTransactionByBlockHashAndIndex` in the helios-ts provider.
- Add info on `eth_getTransactionByHash` and `eth_getTransactionByBlockHashAndIndex` in rpc.md

PR #440, opened 2024-11-21
fix: missing parentheses in helios-ts/lib.ts

eshaan7:
This fixes a typo (missing `{`) introduced in https://github.com/a16z/helios/pull/425.


ncitron (approved):
LGTM!

PR #448, opened 2024-11-25
feat: track filters in state for eth_getFilterChanges

### LINKED ISSUE #447: [bug] eth_getFilterChanges only supports log return type
From alloy-rpc-types:

```rust
pub enum FilterChanges<T = Transaction> {
    /// Empty result.
    #[serde(with = "empty_array")]
    #[default]
    Empty,
    /// New logs.
    Logs(Vec<RpcLog>),
    /// New hashes (block or transactions).
    Hashes(Vec<B256>),
    /// New transactions.
    Transactions(Vec<T>),
}
```

The `eth_getFilterChanges` can return either of the 3 types but the current code only handles `Vec<Log>`. So if I pass a filter ID returned from `eth_newBlockFilter` or `eth_newPendingTransactionFilter` to Helios' `eth_getFilterChanges` then I get the following error:

```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": 1,
        "message": "rpc error on method: get_filter_changes, message: deserialization error: invalid type: string \"0x34bd63ac759af80513e814a47bb4419ed8b8fff5f84730a0a704f3d3418b77f8\", expected struct Log at line 1 column 69"
    },
    "id": "14"
}
```

eshaan7:
Closes #447.

-----

1. Transaction hashes:

image

2. Block hashes:

image

3. Logs:

image


ncitron (changes requested):
This is a good start, but we want to be verifying all data returned by the filter is valid. The current standard we hold ourselves to is that an RPC should be able to withhold parts of the data, but should not be able to insert additional incorrect data (we'd like to check for exhaustiveness at some point too but there are some unfortunate blockers there for a few of the rpc methods like log fetching).

It looks like we aren't checking validity for the non-log filter return types. We should be able to do this by checking if any returned block, blockhash, transaction, or transaction hash has actually been seen by Helios. All of the data to perform these checks should be accessible from the `State` type.

eshaan7:
> It looks like we aren't checking validity for the non-log filter return types. We should be able to do this by checking if any returned block, blockhash, transaction, or transaction hash has actually been seen by Helios

So the `eth_newBlockFilter` filter returns block hashes, while the `eth_newPendingTransactionFilter` filter returns transaction hashes. In both cases, the response is an array of hashes without any indication of whether they are block hashes or transaction hashes. This makes verification quite tricky because Helios cannot verify the new transaction hashes — these are just transactions in the mempool and have not yet been included in any block. For block hashes, I can check if they exist in the state, but if they don’t, I can’t simply error out because they might actually be transaction hashes, and there’s no way to distinguish between the two.

ncitron:
Oh this is a good point.

How that I think about it we should probably be managing more of the filter logic ourselves rather than farming it out to the RPC.

We should keep track of what filters we have, and only use the RPC to manage log and pending transaction filters. For new block filters we don't really need to use the rpc at all. We can just keep track of those filters in Helios and implement the logic directly for returning all the heads that helios has seen.

eshaan7:
> We can just keep track of those filters in Helios and implement the logic directly for returning all the heads that helios has seen.

That makes sense. We can maintain a list of filter IDs for `eth_newBlockFilter` filter in Helios' state and don't need RPC for it then. However, one small edge case that still remains is that if Helios sees an unknown filter ID, it will fallback to RPC for 3 all filter types. This is, again, because the filter ID or its response does not provide information on the filter type.

ncitron:
If Helios sees an unknown filter ID I think we would want to return an error? All filters that a client asks for should have been previously created through Helios before. The same would be true when using an RPC.

eshaan7:
@ncitron; Implemented tracking filter and changes (for eth_newBlockFilter) in state. PTAL. LMK if you think I missed anything.

core/src/client/rpc.rs
```diff
@@ -124,7 +124,7 @@ trait EthRpc<TX: TransactionResponse + RpcObject, TXR: RpcObject, R: ReceiptResp
     #[method(name = "getLogs")]
     async fn get_logs(&self, filter: Filter) -> Result<Vec<Log>, ErrorObjectOwned>;
     #[method(name = "getFilterChanges")]
-    async fn get_filter_changes(&self, filter_id: U256) -> Result<Vec<Log>, ErrorObjectOwned>;
+    async fn get_filter_changes(&self, filter_id: U256) -> Result<FilterChanges, ErrorObjectOwned>;
```
ncitron, line 127: Should we also add the newPendingTransactionFilter and newBlockFilter methods to rpc?

core/src/execution/mod.rs
```diff
@@ -360,6 +398,8 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
     }
 
     pub async fn uninstall_filter(&self, filter_id: U256) -> Result<bool> {
+        // remove the filter from the state
+        self.state.remove_filter(&filter_id).await;
```
ncitron, line 402: What if it is a log filter? Wouldn't we need to uninstall the filter in the rpc if it is a log filter?

ncitron (commented):
Looking good. Just two small changes then I think we're good to merge.

core/src/client/rpc.rs
eshaan7: Those two are already part of the RPC. 

https://github.com/eshaan7/helios/blob/master/core/src/client/rpc.rs#L134-L137

eshaan7: The methods are there in the RPC but their corresponding Wasm bindings are missing, for which I had already created an issue (https://github.com/a16z/helios/issues/451).

core/src/execution/mod.rs
eshaan7: Yes, that's what the next line (L403) does. Regardless of log type, we always call both `state.remove_filter` and `rpc.uninstall_filter`.

https://github.com/eshaan7/helios/blob/d01f43245c87e07222cde003cd8f57348b8dad8e/core/src/execution/mod.rs#L403

PR #452, opened 2024-11-26
feat: added eth_BlobBaseFee

varun-doshi:
Ref #200 

Added `eth_blobBaseFee`

### DIFF
--- a/core/src/client/rpc.rs
+++ b/core/src/client/rpc.rs
@@ -84,6 +84,8 @@ trait EthRpc<TX: TransactionResponse + RpcObject, TXR: RpcObject, R: ReceiptResp
     async fn gas_price(&self) -> Result<U256, ErrorObjectOwned>;
     #[method(name = "maxPriorityFeePerGas")]
     async fn max_priority_fee_per_gas(&self) -> Result<U256, ErrorObjectOwned>;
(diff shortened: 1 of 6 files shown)


core/src/client/rpc.rs
```diff
@@ -84,6 +84,8 @@ trait EthRpc<TX: TransactionResponse + RpcObject, TXR: RpcObject, R: ReceiptResp
     async fn gas_price(&self) -> Result<U256, ErrorObjectOwned>;
     #[method(name = "maxPriorityFeePerGas")]
     async fn max_priority_fee_per_gas(&self) -> Result<U256, ErrorObjectOwned>;
+    #[method(name = "getBlobBaseFee")]
```
eshaan7, line 87: As per the [spec](https://ethereum.github.io/execution-apis/api-documentation/), it is `eth_blobBaseFee` so the `get` is not required.

ncitron (changes requested):
Also there are some clippy and fmt errors introduced here related to unused imports.

eshaan7:
nit: need to update method name in PR title as well.

core/src/client/rpc.rs
ncitron, line 87:
```suggestion
    #[method(name = "blobBaseFee")]
```

PR #453, opened 2024-11-27
Typo Update lib.rs

mdqst:
### Description:
I noticed a small but important correction in the documentation. The phrase **"Examples on how you can use helios"** sounded a bit awkward and could be phrased more clearly. I’ve updated it to **"Examples of using Helios"**.

### Why it matters:
The original wording could cause slight confusion or awkwardness for readers. By adjusting the phrasing, the sentence now reads more naturally and aligns better with common English expression, improving readability and clarity for users.


ncitron:
We are not accepting these kinds of changes at this time.

PR #467, opened 2024-12-09
Fix incorrect testing command in README.md

mdqst:
### Description  
This change corrects a minor but important typo in the documentation under the **Testing** section. The command for running all tests was written as:  

Снимок экрана 2024-12-09 в 15 00 56

However, this is not a valid Cargo command. The correct syntax is:  

```sh
cargo test --all
```  

This fix ensures that users following the documentation can successfully run all tests without confusion or errors. The previous version of the command could lead to failed execution, potentially frustrating contributors or users attempting to validate the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ cp .env.example .env
 To run all tests, use the following command:
 
 ```sh
-cargo test-all
+cargo test --all
 ```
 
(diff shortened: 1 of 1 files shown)


ncitron:
This command is defined in `.cargo` and is valid.

PR #468, opened 2024-12-10
feat: Allow `opstack` to be used as a library

kuksag:
Rationale: I needed to run Helios with OPStack in my application but realized that it's not possible if the dependency on Helios is added as a Git submodule.
This MR exposes the `opstack` module to be used as a library, similar to the `core` and `ethereum` modules


ncitron (approved):
LGTM!

PR #471, opened 2024-12-14
Align Method Examples with Rust Documentation Standards: Remove &self

futreall:
### Why Remove `&self` in Documentation Examples?

Including `&self` in method examples violates **Rust's documentation conventions** and adds unnecessary clutter. It may confuse users into thinking `&self` needs to be explicitly passed when calling methods, which is not the case. Omitting `&self` makes the examples cleaner, easier to read, and aligns with standard practices.


ncitron:

PR #472, opened 2024-12-14
Update README.md

detrina:
Description:
This pull request addresses a couple of minor errors in the documentation:

"documentation of" → "documentation for"
"like so" → "as follows" (Replaced "like so" with "as follows" for more formal technical language.)
I hope this correction is helpful, and I will continue to contribute in a positive way. Thank you!


ncitron:

PR #479, opened 2024-12-22
fix: several typos of various value

CreeptoGengar:


ncitron:

Issue #474, opened 2024-12-17
refactor: use typed url instead of strings


ncitron:
There are many places where we use string to represent urls instead of the `Url` type. This causes strange behavior when we try to build urls such as trailing slashes breaking things. We should take a look throughout the codebase to ensure we are using these strongly typed Urls that support safe building of endpoints instead.

SkandaBhat:
I can look into this

PR #480, opened 2024-12-23
refactor: change rpc from string to url in cli entrypoint

SkandaBhat:
partially towards #474 

replaces string type to url in cli. There are other places this type change can be implemented, will raise similar small PRs towards #474


ncitron (approved):
LGTM!

PR #481, opened 2024-12-27
fix: add WASM-compatible retry mechanism for RPC requests

### LINKED ISSUE #407: feat: support retries when execution rpc fails in wasm
We had to disable Alloy's rate limit retry mechanism when compiling to wasm in `core/execution/rpc/http_rpc.rs` because it doesn't seem to compile properly. This means if the external RPC fails the whole helios request will error, which we don't want. Implement some sort of retry mechanism that works in wasm or find out how to make Alloy's work properly (using alloy's is the preference). We should also be intelligent about only retrying things that we think might be resolved (rate limits, connection errors) as opposed to something else.

DeVikingMark:
Fixes #407

Added WASM-compatible retry mechanism for RPC requests to replace non-working Alloy retry implementation.

Changes made:
- Implemented custom retry logic using wasmtimer::tokio::sleep
- Added RetryConfig for configurable retry parameters
- Added exponential backoff with configurable initial and max delays
- Added proper error handling for different failure types
- Added comprehensive tests using mockito

This fixes the issue where RPC retries would fail when compiled to WebAssembly.


core/execution/rpc/http_rpc.rs
```diff
@@ -0,0 +1,76 @@
+use std::time::Duration;
```
ncitron, line 1: It seems that this is not in the right directory so isn't even being compiled? Should be in src/execution/rpc.

ncitron:
This has been resolved as of #484 since alloy's retries work in wasm as of the latest updates.

Issue #484, opened 2024-12-31
Update alloy dependency version


karen-sarkisyan:
Update version of `alloy` and `op-alloy` crates in dependencies.

As discussed in dev chat, it might be beneficial for issues #407 and #302 , so we might want to do it before alloy goes stable.

PR #485, opened 2024-12-31
chore: update alloy version to 9.0

karen-sarkisyan:
This is my take on issue #484 . I tried to keep it as simple as possible without any rework of underlying structure.

Key changes:
`alloy` crates have been updated to latest version of `9.0`

Some breaking changes are likely to be introduced because of change `TransactionRequest` -> `OpTransactionRequest` type for OPStack.

I invite anyone interested to review and regression test this PR, feedback is much appreciated. I'd be happy to fix whatever has to be fixed.

### DIFF
--- a/opstack/src/consensus.rs
+++ b/opstack/src/consensus.rs
@@ -1,10 +1,12 @@
 use alloy::consensus::Transaction as TxTrait;
 use alloy::primitives::{b256, fixed_bytes, keccak256, Address, B256, U256, U64};
 use alloy::rlp::Decodable;
-use alloy::rpc::types::{EIP1186AccountProofResponse, Parity, Signature, Transaction};
+use alloy::rpc::types::EIP1186AccountProofResponse;
+use alloy::rpc::types::Transaction as EthTransaction;
 use alloy_rlp::encode;
 use eyre::{eyre, OptionExt, Result};
 use op_alloy_consensus::OpTxEnvelope;
+use op_alloy_rpc_types::Transaction;
 use std::str::FromStr;
 use std::time::Duration;
 use tokio::sync::mpsc::Sender;
@@ -215,7 +217,7 @@ fn verify_unsafe_signer(config: Config, signer: Arc<Mutex<Address>>) {
 
             // with storage proof
             let storage_proof = proof.storage_proof[0].clone();
-            let key = storage_proof.key.0;
+            let key = storage_proof.key.as_b256();
             if key != B256::from_str(UNSAFE_SIGNER_SLOT)? {
                 warn!(target: "helios::opstack", "account proof invalid");
                 return Err(eyre!("account proof invalid"));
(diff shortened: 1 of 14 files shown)


opstack/src/consensus.rs
```diff
@@ -268,140 +270,65 @@ fn payload_to_block(value: ExecutionPayload) -> Result<Block<Transaction>> {
...
-                    };
-
-                    tx
+                    inner_tx.from = inner.inner().from;
+                    let deposit_nonce = Some(inner.inner().nonce());
+
+                    Transaction {
+                        inner: inner_tx,
+                        deposit_nonce,
+                        deposit_receipt_version: None,
```
karen-sarkisyan, line 330: I've left it as `None` for now. Or do we want to implement a check for Canyon fork in scope of this PR?

ncitron: I think we should assume that Canyon has been activated for now as all the superchain chains have already gone through that hard fork. Alternatively we can add the canyon time to the config so we can be smart about what to set this field.

Either way, I think setting it to None as the default is probably not the right choice as Canyon is already enabled.

opstack/src/spec.rs
```diff
@@ -133,162 +139,174 @@ impl NetworkSpec for OpStack {
...
     fn can_build(&self) -> bool {
-        TransactionBuilder::<Ethereum>::can_build(self)
+        <TransactionRequest as TransactionBuilder<Ethereum>>::can_build(self.as_ref())
     }
 
     #[doc(alias = "output_transaction_type")]
     fn output_tx_type(&self) -> OpTxType {
-        OpTxType::try_from(self.preferred_type() as u8).unwrap()
+        match self.as_ref().preferred_type() {
+            TxType::Eip1559 | TxType::Eip4844 => OpTxType::Eip1559,
```
ncitron, line 287: Is the thought here that we should never see a EIP-4844 type here so we can just map it to 1559 to keep the function infallible?

ncitron (commented):
Looking good. Just one thought on the Canyon stuff.

opstack/src/spec.rs
karen-sarkisyan: Yes. But full disclosure: I stole the implementation for `TransactionBuilder` from [this PR in op-alloy](https://github.com/alloy-rs/op-alloy/pull/215). So I can't say I'm sure that converting without fail is the best way for us, but that's the way alloy team decided to go.

opstack/src/consensus.rs
karen-sarkisyan: Maybe I went the wrong way here.

I used [op-alloy's Transaction struct](https://docs.rs/op-alloy/0.9.0/op_alloy/rpc_types/struct.Transaction.html) that has `deposit_nonce` and `deposit_receipt_version` fields.
While `deposit_receipt_version` would be easy with known block number, I think I used a wrong nonce -- the correct one is in the receipt, not envelope. And I'm not sure how to get receipt's data here.

Maybe this is why you used the [alloy's generic Transaction](https://docs.rs/alloy-rpc-types-eth/0.9.0/alloy_rpc_types_eth/transaction/struct.Transaction.html) in the first place and we should stick with it? What do you think?

ncitron: Hmm it seems there actually isn't even a way to extract the correct deposit nonce anyway, which is pretty unfortunate. I guess we can leave as is for now.

PR #489, opened 2025-01-13
feat(consensus): implement EIP-4844 transaction handling

VolodymyrBg:
Commit Description:
Replace todo!() with proper implementation of EIP-4844 transaction handling in consensus.rs. This change adds full support for blob transactions (EIP-4844) including:
- Signature recovery and verification
- Access list handling
- Fee parameters (max_fee_per_gas, max_priority_fee_per_gas, max_fee_per_blob_gas)
- Blob versioned hashes
- Gas price calculation based on base fee


ncitron:
By decoding the tx envelope from RLP this is already handled automatically. I have confirmed that Helios currently fully decodes EIP-4844 transactions fully. Is there anything Helios doesn't seem to be decoding that you have observed?

I do really appreciate the effort though.

Issue #496, opened 2025-01-23
fix: demo app for helios-ts fails to display blocks data


karen-sarkisyan:
Steps to reproduce:
1. Compile helios-ts by running `npm run build` in helios-ts directory 
2. In helios-ts/index.html add your API key and adjust configs if necessary
3. Open helios-ts/index.html in your browser

Expected result: app displays latest blocks data

Actual result: no blocks are displayed, errors in the console

PR #497, opened 2025-01-23
Fix: process Map into an Object for blocks in helios-ts

karen-sarkisyan:
fixes: #496

After latest changes, `Block` is serialized differently which results in `Map` in JavaScript. This is not an expected type from a provider.

This PR uses existing `mapToObj` function to process the `Map` into an `Object`.


ncitron (approved):
LGTM!

PR #499, opened 2025-01-26
chore: Fix incorrect variable usage in println! Update checkpoints.rs

famouswizard:
I noticed that the `println!` macro was missing some required arguments for the format string.
This caused a compilation error.

I've fixed it by ensuring all variables referenced in the format string are properly passed as arguments.


ncitron:
The original syntax a valid usage of the println macro.

PR #500, opened 2025-01-26
fix: Improve error handling by using String(err) for safer conversion…

voronor:
I updated the error handling in the catch block to use `String(err)` instead of `err.toString()`.
While this isn't a bug fix, it's an improvement to ensure consistent and reliable string conversion, especially in cases where `err` might not be an object.


ncitron:
This prevents custom string printing defined by the error type. Additionally the errors returned should not be nullable and have a toString method anyway.

PR #502, opened 2025-01-31
refactor: custom EVM support

### LINKED ISSUE #406: feat: better evm handling
Right now Helios always uses the same EVM (revm). We should be a bit more generic about this to support any evm, and should properly handle setting up revm differently for Ethereum vs OP Stack. We should also properly initialize the evm differently depending on what hard fork is active.

The best spot to put this functionality is probably as a part of `NetworkSpec`.

Dhruv-2003:
closes #406 

## Description 

This PR adds a new method on the `network_spec` called `evm()` to retrieve the evm with custom handlers and configuration. The evm build is now handled on the respective network implementation that enables support for evms based on the active hard fork. 

## Note

Basic implementation of the `evm` method on the network is added currently, but should be modified to suit our requirements before merging

### DIFF
--- a/core/src/client/node.rs
+++ b/core/src/client/node.rs
@@ -46,16 +46,22 @@ impl<N: NetworkSpec, C: Consensus<N::BlockResponse>> Node<N, C> {
     ) -> Result<Bytes, ClientError> {
         self.check_blocktag_age(&block).await?;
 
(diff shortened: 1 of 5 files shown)


Dhruv-2003:
closing in favour of an updated PR #597

PR #505, opened 2025-02-05
feat: verifiable-api

eshaan7:
### Endpoints

- [x] `eth_getProof` - `GET /eth/v1/proof/account/{address}`
- [x] `eth_getBalance` - `GET /eth/v1/proof/account/{address}`
- [x] `eth_getTransactionCount` - `GET /eth/v1/proof/account/{address}`
- [x] `eth_getCode` - `GET /eth/v1/proof/account/{address}?includeCode=true`
- [x] `eth_getStorageAt` - `GET /eth/v1/proof/account/{address}?storageSlots={slots}`
- [x] `eth_getTransactionReceipt`  - `GET /eth/v1/proof/transaction/{hash}/receipt`
- [x] `eth_getLogs` - `GET /eth/v1/proof/logs`
- [x] `eth_getFilterLogs` - `GET /eth/v1/proof/filterLogs`
- [x] `eth_getFilterChanges` - `GET /eth/v1/proof/filterChanges`
- [x] `eth_call` - `POST /eth/v1/proof/createExtendedAccessList`
- [x] `eth_estimateGas` - `POST /eth/v1/proof/createExtendedAccessList`
- [x] `eth_createAccessList` - `POST /eth/v1/proof/createExtendedAccessList`
- [x] `eth_chainId` -> `GET /eth/v1/chainId`
- [x] `eth_sendRawTransaction` -> `POST /eth/v1/sendRawTransaction`
- [x] `eth_getBlockByHash` -> `GET /eth/v1/block/{block_id}` 
- [x] `eth_getBlockByNumber` -> `GET /eth/v1/block/{block_id}` 
- [x] `eth_getBlockReceipts` -> `GET /eth/v1/block/{block_id}/receipts`
- [x] `eth_newFilter` -> `POST /eth/v1/filter`
- [x] `eth_newBlockFilter` -> `POST /eth/v1/filter`
- [x] `eth_newPendingTransactionFilter` -> `/POST eth/v1/filter`
- [x] `eth_uninstallFilter`  -> `DELETE /eth/v1/filter/{filter_id}`

### Other

- [x] Integration with helios-ts (Wasm)
- [x] Unit tests for verifiable-api
- [x] Integration tests (something similar to RPC equivalent tests)
- [ ] Documentation
- [ ] Dockerfiles


verifiable-api/server/bin/handlers.rs
```diff
@@ -0,0 +1,439 @@
...
+        .map_err(map_server_err)?;
+
+    Ok(Json(AccountResponse {
+        account: Account {
+            balance: proof.balance,
+            nonce: proof.nonce,
+            code_hash: proof.code_hash,
+            storage_root: proof.storage_hash,
+        },
+        code: code.into(),
```
ncitron, line 170: What if we don't care about the code (for example, in the case we are fetching a balance). We should be able to replace it with a codehash in those instances.

verifiable-api/server/bin/handlers.rs
```diff
@@ -0,0 +1,439 @@
...
+        receipt_proofs,
+    }))
+}
+
+/// This method returns an array of all logs matching the filter with given id.
+/// Corresponding to each log, it also returns the transaction receipt and a Merkle proof of its inclusion.
+///
+/// Replaces the `eth_getFilterLogs` RPC method.
+pub async fn get_filter_logs<N: NetworkSpec, R: ExecutionRpc<N>>(
+    Path(filter_id): Path<U256>,
```
ncitron, line 242: What if our rpc doesn't have the filter id? Maybe we could try handling filter ID in Helios and convert it into some representation that doesn't have it? Just thinking about some world where this thing is registering tons of filters to its execution client, but then the client gets reset or something.

There's probably a few ways we could try and handle it.

verifiable-api/server/bin/handlers.rs
```diff
@@ -0,0 +1,439 @@
...
+            txs.into_iter().map(|t| t.inner.tx_hash().clone()).collect(),
+        ),
+    }))
+}
+
+/// This method returns a list of all accounts (along with their proofs) that are accessed by a given transaction.
+/// The list includes the `from`, `to` and `beneficiary` addresses as well.
+///
+/// Replaces the `eth_createAccessList` RPC method.
+pub async fn create_access_list<N: NetworkSpec, R: ExecutionRpc<N>>(
```
ncitron, line 300: One of the reasons we want this API is that eth_createAccessList doesn't work in some cases.

A better way to do this would be to simulate the call in revm and collect all accounts accessed instead of using eth_createAccessList.

verifiable-api/server/bin/handlers.rs
```diff
@@ -0,0 +1,439 @@
...
+            logs.len(),
+            block_nums.len(),
+            MAX_SUPPORTED_BLOCKS_TO_PROVE_FOR_LOGS,
+        )
+        .into());
+    }
+
+    // Collect all tx receipts for all block numbers
+    let blocks_receipts_fut = block_nums.into_iter().map(|block_num| {
+        let rpc = Arc::clone(&rpc);
```
ncitron, line 397: nit: `rpc.clone()`

verifiable-api/types/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+use serde::{Deserialize, Serialize};
+
+use helios_common::network_spec::NetworkSpec;
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AccountResponse {
+    pub account: TrieAccount,
+    pub code: Bytes,
+    pub account_proof: Vec<Bytes>,
```
ncitron, line 18: Why is the account proof just bytes instead of a more structured type like we see in the rpc responses?

verifiable-api/types/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub storage_proof: Vec<EIP1186StorageProof>,
```
ncitron, line 19: nit: `storage_proofs`

verifiable-api/types/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StorageAtResponse {
```
ncitron, line 24: Is this used anywhere?

verifiable-api/types/src/lib.rs
ncitron: Not necessarily against this if it is to ensure the response is small. But this brings the next question of why we aren't doing this for storage proofs.

common/src/types.rs
```diff
@@ -49,3 +65,27 @@ impl<'de> Deserialize<'de> for BlockTag {
         Ok(block_tag)
     }
 }
+
+impl From<BlockTag> for BlockId {
```
ncitron, line 69: Why do we have both `BlockTag` and `BlockId`? Can we use just one of them throughout the codebase?

core/src/client/mod.rs
```diff
@@ -29,11 +28,12 @@ pub struct Client<N: NetworkSpec, C: Consensus<N::BlockResponse>> {
 impl<N: NetworkSpec, C: Consensus<N::BlockResponse>> Client<N, C> {
     pub fn new(
         execution_rpc: &str,
+        verifiable_api: Option<&str>,
```
ncitron, line 31: I don't love that we have both the execution rpc and verifiable api address as inputs. I wonder if there is some way to pass one or the other?

Maybe an enum or alternatively we can construct an the api/rpc client before and pass it in?

core/src/client/node.rs
```diff
@@ -6,27 +6,28 @@ use alloy::primitives::{Address, Bytes, B256, U256};
...
 use crate::execution::ExecutionClient;
-use crate::fork_schedule::ForkSchedule;
-use crate::network_spec::NetworkSpec;
 use crate::time::{SystemTime, UNIX_EPOCH};
-use crate::types::BlockTag;
 
 pub struct Node<N: NetworkSpec, C: Consensus<N::BlockResponse>> {
     pub consensus: C,
-    pub execution: Arc<ExecutionClient<N, HttpRpc<N>>>,
+    pub execution: Arc<ExecutionClient<N, HttpRpc<N>, VerifiableApiClient>>,
```
ncitron, line 22: Similar concern here. Can we have something that encompasses that it would use one or the other rather than both the HttpRpc and VerifiableApi client?

core/src/execution/constants.rs
```diff
@@ -1,5 +1,5 @@
 pub const PARALLEL_QUERY_BATCH_SIZE: usize = 20;
 
-// We currently limit the max number of logs to fetch,
+// We currently limit the max number of blocks to prove a list of logs,
 // to avoid blocking the client for too long.
-pub const MAX_SUPPORTED_LOGS_NUMBER: usize = 5;
+pub const MAX_SUPPORTED_BLOCKS_TO_PROVE_FOR_LOGS: usize = 5;
```
ncitron, line 5: What do you think the right number here is?

ncitron (commented):
Took a first pass. Looking good but have a few comments.

One thing I notice a lot here is that there is always the execution rpc and optionally the verifiable api being passed around, but I think it makes more sense to run it in either rpc or api mode. This will be nice because you could imagine some third way of getting data (like the portal network), and as these options grow things can become painful.

core/src/execution/constants.rs
ncitron: It might also make sense to have different values here for the API vs RPC since the overheads are pretty different for these. The API should be able to handle a higher value.

eshaan7:
> One thing I notice a lot here is that there is always the execution rpc and optionally the verifiable api being passed around, but I think it makes more sense to run it in either rpc or api mode.

So even in API mode, we need the RPC for most JSON-RPC methods that are not implemented in the API. That's why the implementation is such as that the `execution_rpc` is required but you can optionally pass in `verifiable_api` which just switches some methods to use the API instead.

verifiable-api/server/bin/handlers.rs
eshaan7: So my idea was to basically replace the `eth_getProof`, `eth_getTransactionCount`, `eth_getBalance` `eth_getCode` and `eth_getStorageAt` methods with a single endpoint that is `/account/<address>` here.

Currently in Helios, methods like `eth_getBalance`, `eth_getTransactionCount`, `eth_getStorageAt` also fetch `eth_getProof` and `eth_getCode` to verify the account proof.

A single API that returns all required account related data makes it more clean IMHO.

verifiable-api/client/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        block: Option<BlockId>,
+    ) -> Result<AccountResponse>;
+    async fn get_transaction_receipt(
+        &self,
+        tx_hash: B256,
+    ) -> Result<Option<TransactionReceiptResponse<N>>>;
+    async fn get_logs(&self, filter: &Filter) -> Result<LogsResponse<N>>;
+    async fn get_filter_logs(&self, filter_id: U256) -> Result<FilterLogsResponse<N>>;
+    async fn get_filter_changes(&self, filter_id: U256) -> Result<FilterChangesResponse<N>>;
+    async fn create_access_list(
```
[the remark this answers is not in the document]
eshaan7: Not so sure about `call` because the API response is just a list of accounts without the EVM call result.

verifiable-api/types/src/lib.rs
eshaan7: The `account_proof` is `Vec<Bytes>` in [Alloy](https://github.com/alloy-rs/alloy/blob/79964358d42ff4f27d3cb21570dd8026c01d1e75/crates/rpc-types-eth/src/account.rs#L60-L61) as well (for the response of `eth_getProof`).

So it is the same.

verifiable-api/types/src/lib.rs
eshaan7: Yeah, ideally it should be `storage_proofs` but I went with `storage_proof` because that is what is used in JSON-RPC -- see [here](https://github.com/alloy-rs/alloy/blob/79964358d42ff4f27d3cb21570dd8026c01d1e75/crates/rpc-types-eth/src/account.rs#L62-L63).

core/src/execution/constants.rs
eshaan7: Having different values for API vs RPC does make sense. Might also wanna do it for `PARALLEL_QUERY_BATCH_SIZE`.

As for what is the right number, I think we need to make this value configurable by the end-user. For example, somebody using a dedicated RPC could afford a lot more requests.

common/src/types.rs
eshaan7: `BlockTag` is a custom type used for the block param in Helios execution RPC interface. It is restricted to `latest`/`finalized`/`<number>` to facilitate fetching a block from Helios state.

Whereas `BlockId` comes from Alloy and has values like `<hash>`/`pending`/`safe`/`earliest` (in addition to the ones above) and is used by alloy's RPC provider interface.

To use only `BlockId` throughout the codebase, we would need to handle the `pending`/`safe`/`earliest` cases somehow.

Could create a separate issue for this, might be out of scope for this PR since `BlockTag` already exists and is not a new addition in this PR.

core/src/client/node.rs
eshaan7: `VerifiableApi` doesn't completely replace `HttpRpc` (as I said [here](https://github.com/a16z/helios/pull/505#issuecomment-2670692580)). 

So here the `ExecutionClient` is what encapsulates both `VerifiableApi` and `HttpRpc`.

core/src/client/mod.rs
eshaan7: I think that's only possible if passing verifiable_api means you don't need to pass execution_rpc. But that is not the case in the current implementation since the former doesn't completely replace the latter.

ncitron: What methods do we need to fall back on the execution rpc to verify? We should just have versions of those added to the verifiable api (even if they don't really change much from the rpc) so that we can have a single endpoint that helios consumes.

common/src/types.rs
ncitron: Yeah we can leave this be for now.

verifiable-api/server/bin/handlers.rs
ncitron: Yeah exactly. It's harder but `eth_createAccessList` is designed more for simulating a tx so that you can add an access list to your transaction. This means that in many client implementations it is stricter than `eth_call`. For example it checks that the from account has enough eth to pay for gas, which call usually doesn't do. This very often leads to this method failing to return a reasonable access list for our evm call to consume.

verifiable-api/server/bin/handlers.rs
ncitron: It is certainly more clean, but my concern is that contract code can be up to 25kb, which is not an insubstantial amount to waste if we don't need it.

We could have a query parameter that decides whether to include the full code and make the code parameter an optional. We should be sure to document that None does not necessarily mean there is no code at the location, but that we didn't ask for it.

core/src/client/mod.rs
eshaan7: All the methods where response needs to be verified are there in the VerifiableApi (see PR description for list). I've not implemented methods where we just need to proxy the same request/response without any change.

These would include methods like `eth_getBlockReceipts`, `eth_newFilter`, `eth_newBlockFilter` and many more that are used by Helios but also methods like `eth_getBlockTransactionCountByHash`, `eth_getBlockTransactionCountByNumber` and many more which are not used by Helios and rather constructed from Helios state.

We can think about implementing the ones that are used to replace the RPC but I don't see any point in the ones that are not used at all.

verifiable-api/client/src/lib.rs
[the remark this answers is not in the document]
ncitron: Any other ideas on naming? Since we are going to migrate away from the access list mechanism for this method we should probably have some different name to avoid confusion.

verifiable-api/types/src/lib.rs
ncitron: Oh my bad I thought it was just bytes and not a vec of bytes which confused me. This makes sense.

verifiable-api/types/src/lib.rs
ncitron: Since this is a new api might as well change it to have it make a bit more sense.

core/src/client/mod.rs
ncitron: Yeah I agree no need to implement unused methods, but we should ensure that helios can consume either data source as its sole data source rather than needing to rely on both. I think it will only require adding one or two more endpoints.

verifiable-api/server/bin/handlers.rs
ncitron: I'm mostly wondering if there is some improvement we can make to try and avoid the case where the rpc doesn't have the filter installed. Not sure if there is anything better we could do but lets think on it.

verifiable-api/client/src/lib.rs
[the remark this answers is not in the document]
eshaan7: I think if we are changing it to the revm implementation, then we can call it `/simulate-transaction` which returns 3 fields: `output`, `gasUsed`, `accounts`.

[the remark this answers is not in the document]
ncitron: Hmm let me think on it more. Lets leave as is for the time being.

core/src/client/mod.rs
eshaan7: Will need to add the following additional RPC methods:

- `eth_getBlockByHash`
- `eth_getBlockByNumber`
- `eth_chainId`
- `eth_sendRawTransaction`
- `eth_getBlockReceipts`
- `eth_newFilter`
- `eth_newBlockFilter`
- `eth_newPendingTransactionFilter`
- `eth_uninstallFilter`

Shouldn't be too much effort to implement since we just wanna proxy the same request/response.

I'll get started on this?

ncitron: Sounds good

verifiable-api/server/bin/handlers.rs
ncitron: Ah we probably shouldn't also fetch the code when just verifying a balance. The codehash is always correct since it is hashed into the account, which is verified via merkle proof against the stateroot. So this means we don't actually need to fetch the code if we do not care about it.

verifiable-api/server/bin/handlers.rs
eshaan7: One possible improvement can be to remove the filter from Helios' state if RPC returns "filter not found" error. So subsequent requests for the same filter ID won't even hit the RPC.

core/src/execution/constants.rs
eshaan7: Implemented endpoints for these methods in 236a9052564f2832bc0d6780071538511cb4657a.

core/src/client/mod.rs
eshaan7: Implemented endpoints for these methods in 236a9052564f2832bc0d6780071538511cb4657a.

eshaan7: There are two other places in the codebase that directly use the HttpRpc instead of ExecutionClient.

- https://github.com/a16z/helios/blob/b4cabb52732c797f71613d98e1f73ea63cf1aed3/opstack/src/server/mod.rs#L92
- https://github.com/a16z/helios/blob/b4cabb52732c797f71613d98e1f73ea63cf1aed3/core/src/execution/state.rs#L324

I am not sure how we can replace this with ExecutionClient since that also requires `state`.

eshaan7: Need to encapsulate this better if we want the behavior that user only needs to specify either verifiable_api URL or execution_rpc URL but not both.

verifiable-api/server/bin/handlers.rs
eshaan7: > We should be sure to document that None does not necessarily mean there is no code at the location, but that we didn't ask for it.

Could have a query param `?includeCode=<bool>` that omits the `code` field from response if `false`. Don't need to handle the `None` case separately then.

ncitron: Yep I think thats fine. For deserialization purposed you will probably have an easier time having it be an optional in the struct, which when the json omits it deserializes to none.

core/src/client/mod.rs
ncitron: For the server this shouldn't matter as this is backend infra for our opstack integration, and never runs on the client.

As for the state, we should definitely handle that using either the rpc or the api.

verifiable-api/server/bin/handlers.rs
eshaan7: Made the `code` field optional in 77dbd2809f29c638cd3068f09c75e9c3b8eaa604.

core/src/client/mod.rs
eshaan7: This is now solved in 9864ba5da34caa81c2036cfba771f8c0ea37c7b5.

cli/src/main.rs
```diff
@@ -126,8 +127,10 @@ struct EthereumArgs {
     rpc_port: Option<u16>,
     #[clap(short = 'w', long, env)]
     checkpoint: Option<B256>,
-    #[clap(short, long, env, value_parser = parse_url)]
+    #[clap(long, env, value_parser = parse_url)]
```
ncitron, line 130: Lets keep the shorthand for this. It's likely the one people would want to set themselves, since its unlikely there will be other hosts of the verifiable api for some time, so users will likely stick to the default there.

common/src/execution_mode.rs
```diff
@@ -0,0 +1,18 @@
...
+    Rpc(String),
+    VerifiableApi(String),
+}
+
+impl ExecutionMode {
+    pub fn from_urls(rpc: Option<String>, verifiable_api: Option<String>) -> Self {
+        match (rpc, verifiable_api) {
+            (None, Some(verifiable_api)) => Self::VerifiableApi(verifiable_api),
+            (Some(rpc), None) => Self::Rpc(rpc),
+            (Some(_), Some(_)) => {
```
ncitron, line 12: Maybe if both are available default to the verifiable api?

common/src/types.rs
```diff
@@ -0,0 +1,104 @@
...
+};
+use eyre::{eyre, Report, Result};
+use serde::{de::Error, Deserialize, Serialize};
+
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+pub struct Account {
+    pub balance: U256,
+    pub nonce: u64,
+    pub code_hash: B256,
+    pub code: Option<Vec<u8>>,
```
ncitron, line 16: When `Vec<u8>` is serialized/deserialized it is done as an array of numbers. We should use some other type like alloy's `Bytes` type which encodes as a hex string which is a bit more efficient.

ncitron (commented):
Nice changes. I am curious if we can do some more code reuse between the two inner execution clients? Feels like they are doing a lot of the same stuff.

eshaan7:
> I am curious if we can do some more code reuse between the two inner execution clients? Feels like they are doing a lot of the same stuff.

Do you mean the 2 structs: `ExecutionInnerRpcClient` and `ExecutionInnerVerifiableApiClient`, or something else?

common/src/execution_mode.rs
eshaan7: Can do. Maybe could make sense to log a message at warn/info level informing user about it too.

cli/src/main.rs
eshaan7: Oh yes, I made this change to resolve some clippy issue but I think it can be reverted.

common/src/execution_mode.rs
ncitron: I don't think there is a need

ncitron:
> > I am curious if we can do some more code reuse between the two inner execution clients? Feels like they are doing a lot of the same stuff.
> 
> Do you mean the 2 clients as in `ExecutionInnerRpcClient` and `ExecutionInnerVerifiableApiClient`, or something else?

Yep!

cli/src/main.rs
eshaan7:
```bash
Command ethereum: Short option names must be unique for each argument, but '-e' is in use by both 'execution_rpc' and 'execution_verifiable_api'
```

I had removed it because of this error. We can keep `-e` for `execution_rpc` and provide a new shorthand like `-ev` for `execution_verifiable_api`?

eshaan7:
Summary of changes in commit f043d119843594e4d49c5436dc24723cf831d885:
- Added tests for all methods of `ApiService`, `ExecutionInnerVerifiableApiClient`, `ExecutionInnerRpcClient` structs and also for the helper MPT functions in the `proof` file.
- Added a bunch of new mock files in `tests/testdata` and a new `tests/test-utils` crate for code-reuse in all the above.
- Improved `MockRpc`, implementing methods that were previously unimplemented.
- Added `MockVerifiableApi` struct (for usage in tests) for `VerifiableApi` trait -- similar to how we have `MockRpc` struct for `ExecutionRpc` trait.

cc: @ncitron

core/src/execution/client/rpc.rs
```diff
@@ -0,0 +1,609 @@
...
+
+    async fn get_filter_changes(&self, filter_id: U256) -> Result<FilterChanges> {
+        let filter_changes = self.rpc.get_filter_changes(filter_id).await?;
+
+        if filter_changes.is_logs() {
+            let logs = filter_changes.as_logs().unwrap_or(&[]);
+            self.verify_logs(logs).await?;
+        }
+
+        Ok(filter_changes)
```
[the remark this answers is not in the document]
ncitron: Actually looking more deeply at this I notice sometimes the responses from methods in this file are verified, while other times they are not.

I think not having clear responsibility of what things should be verified is suboptimal because it can lead to mistakes in the future. Is there a way we can better separate responsibility between verification and data fetching for both this and the corresponding verifiable api version?

For example block and receipt fetching is done in an unverified manner. Also why do we support fetching chain id and setting a new blocks filter when we don't use those too?

core/src/execution/client/rpc.rs
```diff
@@ -0,0 +1,609 @@
...
+            code,
+            account_proof: proof.account_proof,
+            storage_proof: proof.storage_proof,
+        })
+    }
+
+    /// Verify the integrity of each log entry in the given array of logs by
+    /// checking its inclusion in the corresponding transaction receipt
+    /// and verifying the transaction receipt itself against the block's receipt root.
+    async fn verify_logs(&self, logs: &[Log]) -> Result<()> {
```
ncitron, line 283: A lot of the work here feels similar to what we would need to do in the verifiable api to. Can we get some better code reuse?

core/src/execution/client/verifiable_api.rs
```diff
@@ -0,0 +1,541 @@
...
+        self.verify_logs_and_receipts(&logs, receipt_proofs).await?;
+
+        Ok(logs)
+    }
+
+    async fn get_filter_changes(&self, filter_id: U256) -> Result<FilterChanges> {
+        let filter_changes = self.api.get_filter_changes(filter_id).await?;
+
+        Ok(match filter_changes {
+            FilterChangesResponse::Hashes(hashes) => FilterChanges::Hashes(hashes),
```
ncitron, line 102: Same issue here that we may be returning an unverified blocks response.

core/src/execution/client/verifiable_api.rs
```diff
@@ -0,0 +1,541 @@
...
+        let SendRawTxResponse { hash } = self.api.send_raw_transaction(bytes).await?;
+        Ok(hash)
+    }
+
+    async fn new_filter(&self, filter: &Filter) -> Result<U256> {
+        let NewFilterResponse { id, .. } = self.api.new_filter(filter).await?;
+        Ok(id)
+    }
+
+    async fn new_block_filter(&self) -> Result<U256> {
```
ncitron, line 180: unused?

core/src/execution/mod.rs
```diff
@@ -1,109 +1,68 @@
...
+        Ok(Self {
+            client,
             state,
             fork_schedule,
         })
     }
 
     pub async fn check_rpc(&self, chain_id: u64) -> Result<()> {
-        if self.rpc.chain_id().await? != chain_id {
+        if self.client.chain_id().await? != chain_id {
```
ncitron, line 59: Oh nvm I see why you have chain id in the client. This brings me back to my point on the boundary between verified and unverified data being muddled.

core/src/execution/spec.rs
```diff
@@ -0,0 +1,47 @@
...
+use eyre::Result;
+
+use helios_common::{
+    network_spec::NetworkSpec,
+    types::{Account, BlockTag},
+};
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait ExecutionSpec<N: NetworkSpec>: Send + Sync + 'static {
```
ncitron, line 18: Why implement the execution spec on both the api/rpc clients and on the execution client? Feels like we are overloading the meaning a bit?

core/src/execution/client/verifiable_api.rs
eshaan7: So, the filter is created on the underlying RPC, and its filter ID is stored in Helios and returned to the user. However, when the user requests filter changes for this filter ID, the response is constructed from Helios's state without querying the underlying RPC. This means we can likely eliminate the initial call as well, generating the filter ID on helios side itself.

core/src/execution/spec.rs
eshaan7: So my thought process was like: The impl of `ExecutionSpec` on API/RPC clients is specific to their own nuances (that's why they are inner clients). And then the impl of `ExecutionSpec` on `ExecutionClient` uses one of the inner clients and houses the common logic between the two to avoid duplication.

For example, methods like `get_transaction_receipt`, `get_logs`, `create_access_list`, etc have different verification logic in API vs RPC inner clients so they have different implementations in their own structs but methods like `get_block_receipts` have the same verification logic so that is there in `ExecutionClient`.

I do agree it's not the most straightforward and can lead to confusion and thus faulty code in the future.

Happy to brainstorm ideas around improving this abstraction.

core/src/execution/client/rpc.rs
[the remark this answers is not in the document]
eshaan7: > logs -> verify logs
pending txs -> return txs as given
new block -> error

The `FilterChanges` enum in alloy is defined as:

```rust
pub enum FilterChanges<T = Transaction> {
    /// Empty result.
    #[default]
    Empty,
    /// New logs.
    Logs(Vec<RpcLog>),
    /// New hashes (block or transactions).
    Hashes(Vec<B256>),
    /// New transactions.
    Transactions(Vec<T>),
}
```

The `Transactions` variant is not part of the official JSON-RPC and the `Hashes` variant does not tell us whether these are new block or pending tx hashes. And we do rely on this response for pending tx hashes.

core/src/execution/client/verifiable_api.rs
ncitron: Yeah lets do that

core/src/execution/spec.rs
eshaan7: - Made impl of `get_block_receipts` complete in inner clients.
- Renamed `create_access_list` in inner clients to `create_extended_access_list` to avoid confusion.
- Divided `get_block` into `get_block` and `get_untrusted_block` to avoid confusion.
- Other related changes.

core/src/execution/mod.rs
eshaan7: Left a ToDo for this as discussed.

eshaan7: Let's track it here -- https://github.com/a16z/helios/issues/544.

core/src/execution/client/rpc.rs
eshaan7: Some part of the work is indeed duplicated, yes, but code reuse is not trivial here because in API service case we need `ExecutionRpc.get_block_receipts` whereas in RPC client case we need `ExecutionInnerRpcClient.get_block_receipts` (for verified response). And it's only a small portion of the logic i.e. fetching block receipts for a list of block numbers -- which is the same, not the entire fn.

core/src/execution/client/verifiable_api.rs
eshaan7: So we should get rid of `new_block_filter` from both RPC and API altogether since it will not be used by Helios?

eshaan7:
Note that there's a good amount of redundant code in this PR because of the need to convert b/w `BlockTag` & `BlockId` at many diff places. This would be addressed separately in #524.

core/src/execution/client/verifiable_api.rs
ncitron: Yeah I think if we have unused methods it makes sense to cut them unless you see a reason to keep them?

eshaan7: I think we should keep it in the Verifiable-API for completeness, even if Helios doesn't use it.  

The `POST /eth/v1/filter` endpoint allows requests with a `kind` value of `Logs`, `NewBlocks`, or `NewPendingTransactions`, effectively replacing the `eth_newFilter`, `eth_newBlockFilter`, and `eth_newPendingTransactionFilter` RPC methods.  

Removing `NewBlocks` would make the implementation feel incomplete.

core/src/execution/evm.rs
```diff
@@ -62,20 +60,55 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> Evm<N, R> {
...
             ExecutionResult::Halt { gas_used, .. } => Ok(gas_used),
         }
     }
 
-    async fn call_inner(&mut self, tx: &N::TransactionRequest) -> Result<ResultAndState, EvmError> {
+    pub async fn create_access_list(
+        &mut self,
+        tx: &N::TransactionRequest,
+    ) -> Result<AccessListResultWithAccounts, EvmError> {
+        let (result, accounts) = self.call_inner(tx, false).await?;
```
ncitron, line 76: This is used to create the extended access list in the server, so we should be able to designate whether to use `simulate` as true or false, so that we get the right validation here.

eshaan7: `/createExtendedAccessList` now also accepts `validateTx: bool` in request body to allow control over this. https://github.com/a16z/helios/pull/505/commits/77b838d790f8caa685e8da6a0ea6d74ab51ce81b.

ncitron (approved):

PR #506, opened 2025-02-06
fixes #503: replace custom proof impl and triehash with alloy-trie

### LINKED ISSUE #503: Use alloy-trie crate for MPT
Since helios already depends on alloy for a bunch of primitives, it might make sense to also use [alloy-trie](https://crates.io/crates/alloy-trie) crate for the Merkle Patricia Tree implementation which is maintained by the very same alloy team.

This would replace the custom implementation of [`fn verify_proof`](https://github.com/a16z/helios/blob/016af05ab9c98279c11dccdcb79868515ead1796/core/src/execution/proof.rs#L6) with `alloy_trie::proof::verify_proof` and `triehash_ethereum::ordered_trie_root` with `alloy_trie::root::ordered_trie_root`.

This means we can get rid of `triehash_ethereum` crate as a dependency too -- related #38.

eshaan7:
Closes #503.


ncitron (approved):
LGTM!

PR #507, opened 2025-02-06
fix: ensure logs returned by RPC match filter

eshaan7:


core/src/execution/mod.rs
```diff
@@ -449,6 +465,47 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
...
+            if let Some(to_block) = filter.get_to_block() {
+                if log.block_number.unwrap() > to_block {
+                    return false;
+                }
+            }
+            if !filter.address.matches(&log.address()) {
+                return false;
+            }
+            for (i, topic) in filter.topics.iter().enumerate() {
+                if let Some(log_topic) = log.topics().get(i) {
```
ncitron, line 490: If the log contains no topic at index `i`, but the filter does, then this should be viewed as not matching correct? I think we'd want to have an else arm to this that returns false.

eshaan7:
@ncitron; addressed both comments in https://github.com/a16z/helios/pull/507/commits/ce945c59bc339397594a3e019db90abc09dc6814. PTAL.

ncitron (approved):
LGTM!

PR #510, opened 2025-02-09
fix: correct block hash validation in get_payloads function

avorylli:
The get_payloads function was comparing block_hash() with prev_parent_hash, which is incorrect for validating block chain consistency. Changed to compare parent_hash() with prev_parent_hash to ensure proper block linking.

Impact:
- Fixes potential block synchronization issues
- Ensures correct block chain validation
- Prevents false-positive chain breaks


ncitron:
This function is correct as is. The parent hash of block n must be equal to the block hash of block n-1. This change would cause all calls to `get_payloads` to fail since it would check that the parent hash of block n is equal to the parent hash of block n-1, which it will never be.

PR #513, opened 2025-02-13
fixes #495: param and return types for eth_getStorageAt

### LINKED ISSUE #495: fix: fetching storage slot with less than 32 byte length
In certain cases fetching a slot that is not exactly 32 bytes long fails.

For example this fails:
`cast storage 0xdAC17F958D2ee523a2206206994597C13D831ec7 0x0000000000000000000000000000000000000000000000000000000000000000`

While this succeeds:
`cast storage 0xdAC17F958D2ee523a2206206994597C13D831ec7 0x1000000000000000000000000000000000000000000000000000000000000000`

eshaan7:
Closes #495.

### Changes

- Change `slot` param type to be `U256` instead of `B256`.
- Change return value of `eth_getStorageAt` to be `B256` instead of `U256`.


ncitron (changes requested):
I think it would probably be best to try and mirror the interface of alloy here, which is to make the keys and values both U256. By switching to U256 for the key this should resolve the issue, as U256 is not required to be padded to 32 bytes when deserialized.

As for the return type, alloy uses U256, but this does mean the rpc return type may not be padded with zeros and playing with alchemy it seems responses are usually padded, so maybe the B256 makes more sense, especially if the spec demands padding.

But definitely want to go from JsonStorageKey -> U256 at a minimum, as it is deserialized the same way (I think) and is the more commonly used type.

eshaan7:
> As for the return type, alloy uses U256, but this does mean the rpc return type may not be padded with zeros and playing with alchemy it seems responses are usually padded, so maybe the B256 makes more sense, especially if the spec demands padding.

Yes, I tried with both Geth and Reth and they are padding the return value so that's why I went with `B256`.

eshaan7:
image


@ncitron -- Made the slot param take `U256`. PTAL.

ncitron (approved):
LGTM!

PR #517, opened 2025-02-13
fix: handle empty filter topic in ensure_logs_match_filter

eshaan7:
This fixes a bug introduced in PR #507.

## Changes

- The `ensure_logs_match_filter` fn was not handling the null case for filter topics.
- Removed the unnecessary `async` on `ensure_logs_match_filter` fn.

### DIFF
--- a/core/src/execution/mod.rs
+++ b/core/src/execution/mod.rs
@@ -18,7 +18,7 @@ use crate::types::BlockTag;
 
 use self::constants::MAX_SUPPORTED_LOGS_NUMBER;
 use self::errors::ExecutionError;
-use self::proof::{ensure_logs_match_filter, verify_account_proof, verify_storage_proof};
+use self::proof::{verify_account_proof, verify_storage_proof};
 use self::rpc::ExecutionRpc;
 use self::state::{FilterType, State};
 use self::types::Account;
@@ -280,7 +280,7 @@ impl<N: NetworkSpec, R: ExecutionRpc<N>> ExecutionClient<N, R> {
                 ExecutionError::TooManyLogsToProve(logs.len(), MAX_SUPPORTED_LOGS_NUMBER).into(),
             );
         }
-        ensure_logs_match_filter(&logs, &filter)?;
+        self.ensure_logs_match_filter(&logs, &filter)?;
         self.verify_logs(&logs).await?;
         Ok(logs)
     }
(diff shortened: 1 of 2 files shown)


ncitron:
In `topic.matches` it actually checks if the topic is empty, and if so, considers it a match. This is because an empty topic is effectively a matching topic (the lack of a given topic in a fitler implies that we don't care what it is).

Additionally I think by filtering out empty topics we would be creating a mismatch of indices (although I think since empty topics don't really have a valid serialization it shouldn't matter anyway, but lets supposed someone constructs a filter with an empty topic while calling `getLogs` while using Helios as a library. This would create a mismatch of indices when we filter out the empty topics, leading to us rejecting a valid filter.

For example `[topic0, EMPTY_TOPIC, topic1]` filters to `[topic0, topic1]`, which would match with `[topic0, topic1]`, when we would really want it to match with `[topic0, ANY_TOPIC, topic1]` instead.

Also I think it doesn't make sense to move this into the `Proof` module, as I'd like to keep this for merkle proof logic only.

Feel free to double check my understanding though, as I don't have that deep knowledge about how log filtering works.

eshaan7:
> Additionally I think by filtering out empty topics we would be creating a mismatch of indices

That's a very valid point -- my bad for missing that. Will do a lil more testing.

eshaan7:
@ncitron; Pushed a fix and tried it with `[topic0, null, topic2]`. PTAL.

ncitron:
So I think no change is needed to handle an empty filter since if you look in the code for the `matches` method, it returns `self.is_empty() || self.0.contains(value)` which means our code should handle this already. Additionally as it has been rewritten, it will return that the filter matches as soon as it encounters a single empty topic, which means it won't check for matching any subsequent topics, which is not right.

So I think if I understand correctly, there aren't any code changes needed here (besides maybe moving the check filter method to be a function instead of a method).

eshaan7:
image

If you see the example in the screenshot above, the current logic (in `master`) fails to handle the `[topic0, null, topic2]` case.

eshaan7:
> So I think no change is needed to handle an empty filter since if you look in the code for the matches method, it returns self.is_empty() || self.0.contains(value) which means our code should handle this already.

So this is not working as intended because `filter.topics.iter()` will always yield 4 topics. So even if the request is `[topic0, null, topic2]`, the `Filter` struct makes it `[topic0, null, topic2, null]`. This means when trying to check for a log that only has 3 topics, the code enter the `else` case when no topic is present in the log on index `3` where we return `false` by default.

To avoid this, we need the `is_empty` check on filter topic and basically ignore whatever log topic is on that index. Pushed this fix in https://github.com/a16z/helios/pull/517/commits/5cdeedfb0a99ca03c78cd58f0a589b345953b73f.

ncitron (approved):
Ah this makes sense now.

LGTM!

PR #526, opened 2025-02-21
fix: incorrect naming: verify_unsafe_singer -> verify_unsafe_signer

avorylli:
Fixed a typo in OpStackClientBuilder where "singer" was incorrectly used instead of "signer" in field and method names. This improves code clarity and consistency with cryptographic terminology.


ncitron:
already addressed by #529

Issue #388, opened 2024-10-11
Helios fails to sync (invalid sync committee period)


EzequielPostan:
Hello, I am experimenting with Helios and noticed something that I need to understand and may be a bug.
I would like to use Helios to validate historical data, which means that I am interested in running Helios from the oldest checkpoint possible. Moreover, as time goes by, and more checkpoints are created, I will still want to start from that oldest possible checkpoint.

I started then Helios like this (I am testing changes to the code, so I am building the code)

```
$ cargo run  -- \
      --network mainnet  \
      --consensus-rpc https://www.lightclientdata.org \
      --execution-rpc $EXEC_RPC_WITH_API_KEY
```
I ran it in that way to trigger Helios to use the [BaseConfig checkpoint](https://github.com/a16z/helios/blob/master/config/src/networks.rs#L78) 
```
0xc7fc7b2f4b548bfc9305fa80bc1865ddc6eea4557f0a80507af5dc34db7bd9ce
```
which seems to be for [this block](https://etherscan.io/block/8613568) at slot 8613568 (from Sep-24-2019 07:15:20 PM +UTC) which looks fine. And I know that in theory, I am be able to run with `--checkpoint` and use that checkpoint later to fix that as a starting one.

Now, the issue I have is that Helios fails to sync. This happens at "around" (below is the last update I see validated by Helios in my prints) the point where the finality update has
```
Attested header slot:10152347
Finalized header slot:Some(10152256)
```

Here are the normal logs:
```
$ cargo run  -- \
      --network mainnet  \
      --consensus-rpc https://www.lightclientdata.org \
      --execution-rpc $EXEC_RPC_WITH_API_KEY
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/helios --network mainnet --consensus-rpc 'https://www.lightclientdata.org' --execution-rpc 'https://eth-mainnet.g.alchemy.com/v2/<HERE IS MY API KEY>'`
ConsensusClient RPC: https://www.lightclientdata.org
Default checkpoint leaded 0xc7fc7b2f4b548bfc9305fa80bc1865ddc6eea4557f0a80507af5dc34db7bd9ce: 
2024-10-11T13:08:47.772539Z  INFO helios::rpc: rpc server started at 127.0.0.1:8545
2024-10-11T13:08:48.938388Z  WARN helios::consensus: checkpoint too old, consider using a more recent block
[...]

PR #528, opened 2025-02-22
bug fix: Sync failing for checkpoint > 128

varun-doshi:
Ref #388 

Run the following command to check that the issue is resolved:
```
cargo run ethereum --execution-rpc 'https://eth-mainnet.g.alchemy.com/v2/<API_KEY>' --checkpoint 0xc7fc7b2f4b548bfc9305fa80bc1865ddc6eea4557f0a80507af5dc34db7bd9ce
```

### DIFF
--- a/ethereum/src/consensus.rs
+++ b/ethereum/src/consensus.rs
@@ -363,21 +363,18 @@ impl<S: ConsensusSpec, R: ConsensusRpc<S>> Inner<S, R> {
 
         let mut updates: Vec<Update<S>> = vec![];
         if expected_current_period - bootstrap_period >= 128 {
-            println!("Over 128");
             while bootstrap_period < expected_current_period {
-
                 let batch_size = std::cmp::min(
                     expected_current_period - bootstrap_period,
                     MAX_REQUEST_LIGHT_CLIENT_UPDATES.into(),
                 );
-                println!("Running for {} and batch size:{}",bootstrap_period,batch_size);
                 let update = self
                     .rpc
                     .get_updates(bootstrap_period, batch_size.try_into().unwrap())
                     .await?;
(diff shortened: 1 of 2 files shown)


ethereum/src/consensus.rs
```diff
@@ -357,17 +357,36 @@ impl<S: ConsensusSpec, R: ConsensusRpc<S>> Inner<S, R> {
 
         self.bootstrap(checkpoint).await?;
 
-        let current_period = calc_sync_period::<S>(self.store.finalized_header.beacon().slot);
-        let updates = self
+        let expected_current_slot = self.expected_current_slot();
+        let expected_current_period = calc_sync_period::<S>(expected_current_slot);
+        let mut bootstrap_period = calc_sync_period::<S>(self.store.finalized_header.beacon().slot);
```
ncitron, line 362: Maybe we should rename this since it no longer stays as the bootstrap period the whole time? Maybe `next_updates_fetch_period` or something along those lines?

ethereum/src/consensus.rs
```diff
@@ -357,17 +357,36 @@ impl<S: ConsensusSpec, R: ConsensusRpc<S>> Inner<S, R> {
 
         self.bootstrap(checkpoint).await?;
 
-        let current_period = calc_sync_period::<S>(self.store.finalized_header.beacon().slot);
-        let updates = self
+        let expected_current_slot = self.expected_current_slot();
```
ncitron, line 360: Also this update fetching has gotten a bit complex, so we should probably extract it to a `get_updates` function.

ncitron (commented):
Looks good. Just two comments on style.

ncitron (approved):
LGTM!

PR #534, opened 2025-02-27
add get_block_by_hash & use mapToObj in helios-ts

eshaan7:


helios-ts/lib.ts
```diff
@@ -102,34 +102,41 @@ export class HeliosProvider {
         return this.#client.send_raw_transaction(req.params[0]);
       }
       case "eth_getTransactionReceipt": {
-        return this.#client.get_transaction_receipt(req.params[0]);
-      }
-      case "eth_getTransactionByHash": {
```
eshaan7, line 107: This case was removed because apparently it already exists on L68.

helios-ts/lib.ts
```diff
@@ -150,6 +157,10 @@ export class HeliosProvider {
         const block = await this.#client.get_block_by_number(req.params[0], req.params[1]);
```
ncitron, line 157: Can we fix this? Looks like its accidentally returning blocks.

eshaan7: image

`eth_getBlockByNumber` seems to be fine.

ncitron: Oh my bad the diff made it look like `eth_uninstallFilter` was returning blocks.

ncitron (approved):
LGTM!

PR #535, opened 2025-02-27
fix conflicting State.history_size values

eshaan7:


core/src/execution/constants.rs
```diff
@@ -3,3 +3,5 @@ pub const PARALLEL_QUERY_BATCH_SIZE: usize = 20;
 // We currently limit the max number of logs to fetch,
 // to avoid blocking the client for too long.
 pub const MAX_SUPPORTED_LOGS_NUMBER: usize = 5;
+
+pub const MAX_STATE_HISTORY_LENGTH: usize = 64;
```
eshaan7, line 7: @ncitron; this should be 256 or 64? The previous 64 value was not being used anywhere.

eshaan7: Could also make sense to have different values for Ethereum / OP.

ncitron: 64 is probably fine for now. Maybe would want to make this configurable though in the future.

ncitron (approved):
LGTM!

PR #538, opened 2025-03-01
fix: increase MAX_SUPPORTED_LOGS_NUMBER from 5 to 100

DeVikingMark:
This change increases the maximum number of logs that can be fetched from 5 to 100. 

The current limit is too restrictive for modern dApps that generate more logs during transactions. 

It improves usability while maintaining reasonable performance.


DeVikingMark:
@ncitron I've made some changes, jsut curious what are your thoughts

PR #542, opened 2025-03-04
fix: fix incorrect `B256` conversion for `chain_id`

0xbryer:
I noticed that `B256::left_padding_from(&chain_id.to_be_bytes())` is incorrect because this method doesn't exist in `B256`. It looks like the intention was to convert `chain_id` to a 256-bit value.  

This fixes the issue by using:  

```rust
let chain_id = B256::from_uint(&U256::from(chain_id));
```

Why is this a problem?  
- `B256` expects a 256-bit value, but `chain_id.to_be_bytes()` only returns 8 bytes.


ncitron:
This method exists, and is working as intended.

PR #554, opened 2025-03-07
refactor: avoid duplicate crates

rex4539:


ethereum/consensus-core/Cargo.toml
```diff
@@ -4,15 +4,8 @@ name = "helios-consensus-core"
...
-alloy = { version = "0.9.1", features = [
-    "consensus",
-    "rpc-types",
-    "ssz",
-    "rlp",
-    "k256",
-    "eips",
-] }
-alloy-rlp = "0.3.10"
+alloy = { workspace = true, features = ["consensus", "rpc-types", "ssz", "rlp", "k256", "eips"] }
```
ncitron, line 7: An interesting side effect of using workspace dependencies is that even if we select just these features here, it will inherit all of the default features used in the toplevel alloy import as well, which has several other dependencies.

One of the goals of this crate is to support compiling to zk targets like SP1, RISC0, and Jolt, meaning we want the exact minimum feature flags here to avoid compiler errors caused by importing things like networking libraries.

To resolve, we should either make this a normal import here, or make the top level import have no features flagged, and then flag on just the subset we need in each crate.

ncitron:
Closing as stale. Feel free to reopen if you find fix the flagged issues.

PR #557, opened 2025-03-10
Remove unnecessary assignments and cleanup RLP Encoding 

Pricstas:
- Removed unnecessary `_ =` assignment in `for_each` when adding bootnodes.  
- Removed redundant `.into()` conversion when adding RLP value to the `Enr` builder.  

## Why?  
- `_ =` in `for_each` was redundant and provided no functional benefit.  
- `opstack_data.into()` was unnecessary since `opstack_data` is already `Vec<u8>`.


ncitron:
- The underscore assignment prevents clippy warning about ignoring a `Result` type
- The `into` is required for converting the data from `Vec<u8>` to `Bytes`
- We don't accept these types of PRs regardless due to AI spam

PR #559, opened 2025-03-14
fix(opstack): remove deprecated #[clap] attribute

healthyyyoung:
updates deprecated #[clap(...)] attributes to their modern equivalents in clap 4.x.


ncitron:
`clap` still seems to be supported and is used in many places in the codebase, so want to keep it this way for consistency until we can migrate everything over.

healthyyyoung:
> `clap` still seems to be supported and is used in many places in the codebase, so want to keep it this way for consistency until we can migrate everything over.

maybe we can run `cargo check --features clap/deprecated` and fix all in this PR? @ncitron

PR #561, opened 2025-03-17
fix: fix syntax compatibility with older Rust versions

0xstratus:
Replaced the newer syntax (`let Some(variable) = option else { ... }`) with `if let`, ensuring compatibility with older Rust versions and improving readability.


ncitron:
We are not accepting these types of PRs at this time. If you have an actual project that requires support for these older version of Rust, let me know and we can discuss.

PR #565, opened 2025-03-19
fix: panic when transaction env is built for opstack

### LINKED ISSUE #564: panic when building tx env for opstack
There's a bug in https://github.com/a16z/helios/blob/master/opstack/src/spec.rs#L144 that was introduced in PR #485 that causes a panic when building a transaction for opstack.

karen-sarkisyan:
fixes #564


ncitron (approved):
LGTM!

PR #569, opened 2025-03-21
fix: evm state fetch infinite loop

ncitron:
Some evm executions get stuck in an infinite loop when it needs to fetch slot a and b but does not realize it at the same time. This leads to fetching a, then fetching b (and evicting a from our state cache), leading to fetching a again, and repeating the cycle.


bt-abba:
approved

PR #573, opened 2025-03-26
Remove vendored OpenSSL usage, fix build warnings

strmfos:
**Summary**
- Removes vendored OpenSSL usage, switching to system-provided OpenSSL.
- Fixes Clippy build issues and avoids errors on macOS where `openssl@1.1` is disabled.

**Why It’s Important**
- Homebrew no longer supports `openssl@1.1`, so vendored builds fail.
- Relying on the system’s OpenSSL ensures a clean, up-to-date build environment.


ncitron:
We need the vendored openSSL for wasm support. This issue also doesn't seem to be breaking Helios on MacOS.

PR #575, opened 2025-03-30
fix: panic when transaction env is built for opstack

leopardracer:


ncitron:
Helios does not panic when tx env is built for the opstack.

PR #576, opened 2025-03-31
fix: correct OpenAPI documentation formatting issues

GarmashAlex:
This commit addresses a formatting issue in the OpenAPI documentation:
Removed an improperly formatted empty line in the endpoint description for /eth/v1/proof/filterChanges/{filterId}
Updated OpenAPI version from 3.0.0 to 3.1.0 to match the required pattern


ncitron:
The newline is fine.

If we are doing something that causes 3.0.0 to break lmk, but I think it is fine as of now.

PR #577, opened 2025-04-01
fix: incorrect field usage in `tx_env()`

barajeel:
`tx_env()` sets `max_fee_per_blob_gas`, but it mistakenly assigns the value from `max_fee_per_gas()`.
this seems to be an oversight, as the correct method to use is `max_fee_per_blob_gas()`.


ncitron:
This does not compile.

PR #582, opened 2025-04-08
Update install

Hack666r:


ncitron:
this literally breaks the installer

PR #584, opened 2025-04-15
chore: fix Clippy lints across files

Himess:
This PR fixes various Clippy warnings across the codebase in order to pass `cargo clippy -- -D warnings`.

Erros:

Pasted Graphic 17

Pasted Graphic 18

### Changes:

- Replaced `&PathBuf` with `&Path`
- Removed redundant `.into()` on `u64`
- Removed identity `.map(|(k, v)| (k, v))`
- Converted `assert_eq!(x, true)` to `assert!(x)`
- Replaced field reassignment after `Default::default()` with initializer syntax


ncitron:
Unable to reproduce these errors when running `cargo clippy -- -D warnings` on the latest stable rust version.

PR #586, opened 2025-04-20
chore: updated benchmarks from goerli to sepolia

mw2000:
We fix up benchmarking code and replace benchmarks for goerli with sepolia.

We use the following consensus URL for sepolia:

```
SEPOLIA_CONSENSUS_RPC=http://unstable.sepolia.beacon-api.nimbus.team
```

And update the code to use the weth contract address on sepolia: `0x7b79995e5f793a07bc00c21412e50ecae098e7f9`

Benchmarks can be run with the following command from the repo's root:

```
cargo bench --package helios
```


ncitron (approved):
LGTM!

Issue #587, opened 2025-04-20
Wasm compatibility check is broken


mw2000:
The command to check wasm compatibility when run from the root of the repo fails with errors. We also see that it's disabled in the `test.yml` github action since its specified with `arg` instead of `args`.

Command run:

```
cargo check --target wasm32-unknown-unknown --workspace --exclude helios-cli
```

Error snapshot:

```
   Compiling alloy-sol-macro v0.8.21
    Checking icu_locid v1.5.0
    Checking alloy-rlp v0.3.11
    Checking nybbles v0.3.4
    Checking mio v1.0.3
error[E0432]: unresolved import `crate::sys::IoSourceState`
  --> /Users/mihir/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mio-1.0.3/src/io_source.rs:14:5
   |
14 | use crate::sys::IoSourceState;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ no `IoSourceState` in `sys`

error[E0432]: unresolved import `crate::sys::tcp`
  --> /Users/mihir/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mio-1.0.3/src/net/tcp/listener.rs:19:17
   |
19 | use crate::sys::tcp::{bind, listen, new_for_addr};
   |                 ^^^ could not find `tcp` in `sys`
```

mw2000:
think this is from alloy.. feature gating it should solve the problem

ncitron:
closed in #594

PR #588, opened 2025-04-20
fix: fixed Wasm compatibility check (#587)

gap-editor:
## Description

**Fixed Wasm compatibility check (Issue #587)**

Fixed the Wasm target compatibility check that was failing due to issues with the mio crate as reported in [issue #587](https://github.com/a16z/helios/issues/587). Patched the mio dependency to use a Wasm-compatible fork that properly supports the wasm32-unknown-unknown target. Corrected the GitHub Actions workflow configuration by changing `arg` to `args` in the wasm-compatibility job. This ensures proper CI validation for WebAssembly builds, preventing regressions in Wasm support. Contributors can now rely on the automated checks to verify their changes work correctly across all supported platforms.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,4 +90,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
-          arg: --target wasm32-unknown-unknown --workspace --exclude helios-cli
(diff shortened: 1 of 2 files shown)


ncitron:
This adds a dependency that doesn't exist. Also we want to only compile helios-ts for the compatibility check.

Superseded by #594

PR #597, opened 2025-04-23
refactor: custom VM support

### LINKED ISSUE #406: feat: better evm handling

Dhruv-2003:
closes #406 

## Description
This PR adds new methods on the `network_spec`: `call`, `estimate_gas` & `create_access_list` moved from the evm.rs implementation in the core crate. This changes decouples the execution from the generic EVM interface, allowing each network to define its own EVM implementation.

## Note
- `execution_spec` was moved to common crate from core, for use within `network_spec`
-  Evm implementation added on the specific network, is a modified version of old evm.rs, now including `tx_env` and `block_env` method for that network.


common/src/types.rs
```diff
@@ -143,3 +146,28 @@ pub enum SubscriptionEvent<N: NetworkSpec> {
...
+}
+
+fn display_revert(output: &Option<Bytes>) -> String {
+    match output {
+        Some(bytes) => decode_revert_reason(bytes.as_ref()).unwrap_or(hex::encode(bytes)),
+        None => "execution halted".to_string(),
+    }
+}
+
+pub struct AccessListResultWithAccounts {
```
ncitron, line 170: Doesn't this contain redundant info? The access list result contains accounts already?

ncitron (commented):
I think the main thing we can improve some way to configure revm in the (standard) case that we can just use revm as the evm.

For example, ethereum uses standard revm, linea uses revm on an older evm version, and opstack uses op-revm (part of the revm repo). Is there a way for us to make this general enough that we can have some simple way of configuring revm and using it for all three. This could remove a ton of redundant code.

Another thing we should consider is handling different forks. For example Ethereum is upgrading to pectra soon, so we should consider how to ensure that revm is configured with the correct evm version depending on block number so we can seamlessly transition.

common/src/types.rs
Dhruv-2003: AccessListResult does contain a vector of accounts but it is not in form of a `HashMap<address,account>`, it'd need some level of parsing to retrieve that format and thus I think it might have been kept for `create_extended_access_list` . Used here
https://github.com/a16z/helios/blob/e030480100ff2b21a218eda076a9a80b2f55978c/verifiable-api/server/src/service.rs#L195

Altho the other impl goes on with parsing route on rpc https://github.com/a16z/helios/blob/e030480100ff2b21a218eda076a9a80b2f55978c/core/src/execution/client/rpc.rs#L138

Can keep whatever you feel is correct

Dhruv-2003:
> I think the main thing we can improve some way to configure revm in the (standard) case that we can just use revm as the evm.
> 
> For example, ethereum uses standard revm, linea uses revm on an older evm version, and opstack uses op-revm (part of the revm repo). Is there a way for us to make this general enough that we can have some simple way of configuring revm and using it for all three. This could remove a ton of redundant code.
> 
> Another thing we should consider is handling different forks. For example Ethereum is upgrading to pectra soon, so we should consider how to ensure that revm is configured with the correct evm version depending on block number so we can seamlessly transition.

Yeah, I had similar thought to manage the evm better, I do remember revm having ways to configure the evm, will dive in and share on both of the qns

Dhruv-2003:
I did some research on this. Here are a few things, 

1. Agree that we can remove the generic EVM implementation. Can use standard `revm` for Ethereum & Linea (on london spec). Use `op-revm` for opstack as it specifically handles Optimism's deposit/withdrawal and hardfork logic. Fully generic revm configuration doesn't seem that feasible due to this specialised opstack logic. Some work on simplifying this further is under progress but not ready for use (https://github.com/alloy-rs/evm).
2. Even using the specific `revm` variants above, we still need a custom db layer that handles the modified state (account with storage proofs). Can improve the current ProofDB implementation by implementing `DatabaseAsyncRef` simplifying the db logic during transact. Or we can just keep the current and move it into a shared utility crate.
3. On handling Forks, `revm` manages this well via `CfgEnv.spec_id` which we can determine based on the block. Can use this helper crate here - https://github.com/alloy-rs/hardforks.

opstack/src/builder.rs
```diff
@@ -92,7 +92,16 @@ impl OpStackClientBuilder {
         );
         let consensus = ConsensusClient::new(&config);
         let fork_schedule = ForkSchedule {
-            prague_timestamp: u64::MAX,
+            bedrock_timestamp: 1686068903,
+            regolith_timestamp: 0,
```
ncitron, line 96: Why is regolith enabled at timestamp 0? I think it occurred sometime around bedrock.

opstack/src/builder.rs
```diff
@@ -92,7 +92,16 @@ impl OpStackClientBuilder {
         );
         let consensus = ConsensusClient::new(&config);
         let fork_schedule = ForkSchedule {
```
ncitron, line 94: I know since right now we only have configs for mainnet opstack chains these configs are all identical, but its possible we will have some testnet chains, or even some chains that don't follow the superchain hardfork schedule in the future. We might want to support setting this value from the config then, similar to what we do for Ethereum.

revm-utils/src/evm.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        let spec_id = get_spec_id_for_block(
+            &self.network,
+            &self.fork_schedule,
+            block.header().timestamp(),
+        );
+        let spec_id = spec_id.unwrap_or(SpecId::LATEST);
+        let handler_cfg = HandlerCfg {
+            spec_id,
+            is_optimism: false,
```
ncitron, line 164: This should be `true` when we are on an op chain right?

revm-utils/src/evm.rs
```diff
@@ -0,0 +1,228 @@
...
+            tx: Self::tx_env(tx),
+            block: Self::block_env(&block, &self.fork_schedule),
+            cfg,
+        };
+
+        EnvWithHandlerCfg::new(Box::new(env), handler_cfg)
+    }
+
+    fn tx_env(tx: &N::TransactionRequest) -> TxEnv {
+        // TODO: Handle the Tx fields based on networks
```
ncitron, line 183: We could always move back to generating this in the `NetworkSpec` where we have the required type information. I do think if we can figure out a solution that doesn't require that though it would be preferable.

revm-utils/src/evm.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        BlockEnv {
+            number: U256::from(block.header().number()),
+            coinbase: block.header().beneficiary(),
+            timestamp: U256::from(block.header().timestamp()),
+            gas_limit: U256::from(block.header().gas_limit()),
+            basefee: U256::from(block.header().base_fee_per_gas().unwrap_or(0_u64)),
+            difficulty: block.header().difficulty(),
+            prevrandao: block.header().mix_hash(),
+            blob_excess_gas_and_price: Some(blob_excess_gas_and_price),
```
ncitron, line 225: How does this interact OP networks? Should this be set to `None` in that case?

opstack/src/builder.rs
Dhruv-2003: makes sense, any change required for now ?

ncitron: Yeah I think just for completeness, lets add the fork schedule to the opstack config, and then we can just have a `SuperchainForkSchedule::mainnet` function (and at some point we can add testnets as needed) just like how we have `EthereumForkSchedule::mainnet` in the eth crate.

revm-utils/src/evm.rs
Dhruv-2003: it will be `BlobExcessGasAndPrice::new(0, is_prague)`

revm-utils/src/evm.rs
Dhruv-2003: Kept the fields `None` and left a TODO to be handled later

Dhruv-2003:
closed in favour of #621 , major codebase refactoring

PR #600, opened 2025-05-07
Fix client crate link in verifiable-api README

GarmashAlex:
Changed the broken relative path from verifiable-api/client to ./client to correctly point to the client crate directory from the README location.


ncitron (approved):
LGTM!

PR #602, opened 2025-05-11
chore: pin, align deps; fmt

steph-rs:
#### TL;DR
Pins all workspace dependencies and runs cargo +nightly fmt. Zero functional changes.

#### Motivation
We'd love to use helios inside TEE. But our current approach uses sgx sdk for pretty old version of toolchain - so we need to downgrade a lot of deps. While it's not so hard task and most of the future changes are not relevant to main repo, it would be great to have all dependencies in one place.

Formatted with:
```sh
cargo +nightly fmt -- --check \
    --config imports_granularity=Crate \
    --config group_imports=StdExternalCrate
```
If the formatting config feel heavy-handed, let me know and I'll dial them back.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ helios-ts/helios-*.tgz
 helios-ts/pkg
 
 .vscode
+.idea
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1137,45 +1137,11 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "axum"
-version = "0.7.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
-dependencies = [
- "async-trait",
- "axum-core 0.4.5",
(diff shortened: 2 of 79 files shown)


ig-raiderler:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

ncitron:
Appreciate it!

As for this PR, we probably don't want to change the formatting rules away from the default `cargo fmt`.

For the dependencies, I do have plans to do something like this, but it turns out it is not quite as easy as migrating everything to import from the workspace. This is because flagging any feature in the top-level workspace will cause them to be reflected in any crate that inherits it. This means the simple strategy to just migrate everything over ends up causing problems when building in wasm or for zk provers.

I am likely going to go through and do a very in depth look at how we can do this soon (likely going to go alongside a refactor coming up), but want to be very careful on it, as it touches the dependencies of security critical components such as ethereum-consensus-core.

Appreciate the attempt, and will keep you in the loop as we work on migrating in that direction.

steph-rs:
Okay, got it. But will these issues you mentioned be there if we move all the feature inclusions into the crates themselves instead of the workspace toml? Won't this approach solve all these problems?
I really want to help. If not with this difficult task, then maybe you could list a few github issues that you would like to include in future refactoring?🙏 @ncitron

PR #604, opened 2025-05-13
 chore: bump Alloy, Revm and a few other crates

leruaa:
In the process of bumping SP1 dependencies to `v4.2.0` in sp1-helios, I came accross the following error:

```
error: failed to select a version for `c-kzg`.
    ... required by package `alloy-consensus v0.14.0`
    ... which satisfies dependency `alloy-consensus = "^0.14"` of package `alloy v0.14.0`
    ... which satisfies dependency `alloy = "^0.14.0"` of package `sp1-helios-script v0.1.0 (/home/aurel/dev/sp1-helios/script)`
versions that meet the requirements `^2.1` are: 2.1.1, 2.1.0

the package `c-kzg` links to the native library `ckzg`, but it conflicts with a previous package which links to `ckzg` as well:
package `c-kzg v1.0.3`
    ... which satisfies dependency `c-kzg = "^1.0.3"` of package `revm-precompile v16.2.0`
    ... which satisfies dependency `revm-precompile = "^16.2.0"` of package `revm v19.7.0`
    ... which satisfies dependency `revm = "^19.7.0"` of package `helios-ethereum v0.8.7 (https://github.com/a16z/helios?tag=0.8.7#e0304801)`
    ... which satisfies git dependency `helios-ethereum` of package `sp1-helios-script v0.1.0 (/home/aurel/dev/sp1-helios/script)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the `links = "ckzg"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.
```

The conflict is caused by 2 incompatible versions of Alloy are used, so I updated Alloy to 0.14, and also everything that depends on it (`revm`, `ethereum_ssz`, `ssz_types`...)

### DIFF
--- a/core/src/execution/evm.rs
+++ b/core/src/execution/evm.rs
@@ -328,7 +328,7 @@ impl<N: NetworkSpec> Database for ProofDB<N> {
     }
 
     fn code_by_hash(&mut self, _code_hash: B256) -> Result<Bytecode, DatabaseError> {
-        unimplemented!("should never be called")
+        Err(DatabaseError::Unimplemented)
     }
 }
 
--- a/core/src/execution/evm.rs
+++ b/core/src/execution/evm.rs
@@ -152,6 +152,7 @@ impl<N: NetworkSpec> Evm<N> {
         cfg.disable_block_gas_limit = !validate_tx;
         cfg.disable_eip3607 = !validate_tx;
         cfg.disable_base_fee = !validate_tx;
+        cfg.disable_nonce_check = !validate_tx;
 
         Context::mainnet()
             .with_tx(N::tx_env(tx))
--- a/core/src/execution/evm.rs
+++ b/core/src/execution/evm.rs
@@ -1,16 +1,15 @@
-use std::{borrow::BorrowMut, collections::HashMap, mem, sync::Arc};
+use std::{collections::HashMap, mem, sync::Arc};
 
 use alloy::{
     network::{primitives::HeaderResponse, BlockResponse},
-    rpc::types::{AccessListResult, EIP1186StorageProof},
+    rpc::types::{AccessListItem, AccessListResult, EIP1186StorageProof},
 };
-use eyre::{Report, Result};
+use eyre::Result;
 use revm::{
-    primitives::{
-        address, AccessListItem, AccountInfo, Address, Bytecode, Bytes, CfgEnv, Env,
-        ExecutionResult, B256, U256,
-    },
-    Database, Evm as Revm,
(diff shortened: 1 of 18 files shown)


core/src/execution/evm.rs
```diff
@@ -297,9 +300,9 @@ impl<N: NetworkSpec> EvmState<N> {
 }
 
 impl<N: NetworkSpec> Database for ProofDB<N> {
-    type Error = Report;
+    type Error = DatabaseError;
```
leruaa, line 303: I changed that because the `Error` type needs to impl std `Error` now

ethereum/consensus-core/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version.workspace = true
 edition = "2021"
 
 [dependencies]
-alloy = { version = "0.9.1", features = [
+alloy = { workspace = true, features = [
```
ncitron, line 7: By importing this as a workspace import it means we inherit all of the features that are flagged in the top level workspace. I worry that this will break some usecases like various zk provers (although I guess sp1 doesn't run into an issue here). Still probably worthwhile to leave it as its own import for now just to ensure that this critical piece only flags the minimal number of features.

core/src/execution/evm.rs
```diff
@@ -313,18 +317,18 @@ impl<N: NetworkSpec> Database for ProofDB<N> {
...
-    fn storage(&mut self, address: Address, slot: U256) -> Result<U256, Report> {
+    fn storage(&mut self, address: Address, slot: U256) -> Result<U256, DatabaseError> {
         trace!(target: "helios::evm", "fetch evm state for address={:?}, slot={}", address, slot);
         self.state.get_storage(address, slot)
     }
 
-    fn code_by_hash(&mut self, _code_hash: B256) -> Result<Bytecode, Report> {
-        Err(eyre::eyre!("should never be called"))
+    fn code_by_hash(&mut self, _code_hash: B256) -> Result<Bytecode, DatabaseError> {
+        unimplemented!("should never be called")
```
ncitron, line 331: Just to be extra safe here lets return an error so that if revm's behavior changes Helios doesn't crash, but instead errors. We shouldn't use the StateMissing error though, just to avoid an infinite loop of trying to refetch the state if this triggers.

ncitron (commented):
LGTM! Just some small changes.

ncitron (approved):

PR #605, opened 2025-05-14
Replace deprecated consensus RPC endpoint

Bilogweb3:
Hey! I updated the Helios configuration, swapping the non-working URL `https://www.lightclientdata.org` with the working endpoint `https://lodestar-mainnet.chainsafe.io`.


ncitron:
That endpoint seems to be working just fine (just tested it). We'd prefer to keep this one as the default as we can ensure high uptime.

PR #611, opened 2025-05-23
add link to eip-1186

lechpzn:
added link to [eip-1186](https://eips.ethereum.org/EIPS/eip-1186) as research material.


ncitron:
I don't think we need to add a link here

PR #616, opened 2025-05-29
feat: migrate to Vite, add ESM bundle

karen-sarkisyan:
What this PR does:
- Makes the UMD build of helios work in Node.js (browser support is kept)
- Adds an ESM build of helios compatible with Node & Browser
- Both UMD and ESM bundles are standalone js or mjs files respectively with WASM inlined.

How it does it:
- Switches bundler from Webpack to Vite (which uses Rollup) to inline WASM and bundle cross-platform glue JS code in one file
- Uses `wasm-pack` directly without a plugin for better control of preparing the WASM package
- switches from node's `events` lib to `eventemitter3` for full browser/node compatibility

It maximizes compatibility and portability: each standalone file can be used by both Node and browsers with no dependencies. I expect that that down the road we will expand it with more targeted bundles.


ncitron (approved):
LGTM!

PR #622, opened 2025-06-04
Update rpc.md

otc-png:
Let me explain the specific changes made to rpc.md and their purposes:

1. The main change in rpc.md was to improve the documentation for the `eth_subscribe` method. Here's the before and after:

Before:
```markdown
| `eth_subscribe` | `subscribe` | Subscribes to events. Only "newHeads" is currently supported. | `client.subscribe(&self, event_type: &str)` |
```

After:
```markdown
| `eth_subscribe` | `subscribe` | Subscribes to specified event types. Currently supports "newHeads" event type, which notifies about new block headers being added to the chain. Additional event types may be supported in future versions. | `client.subscribe(&self, event_type: &str)` |
```

This change is particularly important because `eth_subscribe` is a crucial method for developers who need to monitor blockchain state changes in real-time. The improved documentation helps them better understand:  what they can subscribe to (currently only "newHeads"), what information they'll receive (new block headers), that the API might expand in the future


ncitron:
agreed

PR #623, opened 2025-06-05
fix: unsupported place in code

DeVikingMark:
There was a broken link in README.md that led to a non-existent doc fragment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ To run Helios on Ethereum, run the command below, replacing `$ETH_RPC_URL` with 
 helios ethereum --execution-rpc $ETH_RPC_URL
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,7 +27,7 @@ To run Helios on Ethereum, run the command below, replacing `$ETH_RPC_URL` with
 helios ethereum --execution-rpc $ETH_RPC_URL
 ```
 
-`$ETH_RPC_URL` must be a [supported Ethereum Execution API Provider](#supported-execution-api-providers) that provides the `eth_getProof` endpoint. Infura does not currently support this. We recommend using Alchemy.
+`$ETH_RPC_URL` must be a supported Ethereum Execution API Provider that provides the `eth_getProof` endpoint. Infura does not currently support this. We recommend using Alchemy.
```
ncitron, line 30: apparently Infura works now, so we should remove the reference to it not woking.

ncitron (approved):

PR #629, opened 2025-06-12
refactor: use Url type consistently for URL representation

### LINKED ISSUE #474: refactor: use typed url instead of strings

ncitron:
## Summary

This PR refactors Helios to consistently use the `Url` type from the `url` crate for URL representation throughout the codebase, replacing string-based URL handling.

### Core Refactoring
- Changed all URL-related struct fields from `String` to `Url` type
- Updated function signatures to accept `&Url` instead of `&str`
- Modified builders to use `IntoUrl` trait with proper error handling

### Specific Updates
- **Config structures**: Updated `consensus_rpc`, `execution_rpc`, `verifiable_api`, and `fallback` fields to use `Url`
- **ConsensusClient**: Now accepts `&Url` in `new()` method
- **Checkpoints**: `get` function now accepts `&Url`
- **OpStack consensus**: Refactored to use `Url` type internally
- **VerifiableApi trait**: **BREAKING CHANGE** - `new()` method now accepts `&Url` instead of `&str`
- **Builder methods**: **BREAKING CHANGE** - URL-accepting methods now return `Result<Self>` instead of `Self`

### Build Improvements
- Fixed binary name collisions:
  - `opstack/server` → `opstack-server`
  - `verifiable-api/server` → `verifiable-api-server`

### Error Handling
- Builder methods that accept URLs now return `Result<Self>` for proper error propagation
- Replaced `unwrap()` calls with `expect()` for better error messages
- Added proper error propagation in URL joining operations
- Updated MockRpc to handle file:// URLs correctly

### Dependencies
- Added `url` crate dependency to `helios-ts` and `helios-core`

## Breaking Changes
1. **VerifiableApi trait**: `new()` method signature changed from `&str` to `&Url`
2. **Builder methods**: The following methods now return `Result<Self>`:
   - `EthereumClientBuilder::consensus_rpc()`
   - `EthereumClientBuilder::execution_rpc()`
   - `EthereumClientBuilder::verifiable_api()`
   - `EthereumClientBuilder::fallback()`
   - `LineaClientBuilder::execution_rpc()`

## Migration Guide

Before:
```rust
let client = EthereumClientBuilder::new()
    .consensus_rpc("https://example.com")
    .build()?;
```

After:
```rust
let client = EthereumClientBuilder::new()
    .consensus_rpc("https://example.com")?  // Note the ? operator
    .build()?;
```

## Testing
- All tests updated and passing
- Fixed sync test to handle URL types correctly
- Updated all examples to use the new API

Closes #474

### DIFF
--- a/ethereum/src/builder.rs
+++ b/ethereum/src/builder.rs
@@ -73,23 +73,31 @@ impl<DB: Database> EthereumClientBuilder<DB> {
         self
     }
 
-    pub fn consensus_rpc<T: IntoUrl>(mut self, consensus_rpc: T) -> Self {
-        self.consensus_rpc = Some(consensus_rpc.into_url().expect("Invalid consensus RPC URL"));
-        self
+    pub fn consensus_rpc<T: IntoUrl>(mut self, consensus_rpc: T) -> Result<Self> {
+        self.consensus_rpc = Some(
+            consensus_rpc
+                .into_url()
+                .map_err(|_| eyre!("Invalid consensus RPC URL"))?,
+        );
+        Ok(self)
     }
 
-    pub fn execution_rpc<T: IntoUrl>(mut self, execution_rpc: T) -> Self {
-        self.execution_rpc = Some(execution_rpc.into_url().expect("Invalid execution RPC URL"));
-        self
+    pub fn execution_rpc<T: IntoUrl>(mut self, execution_rpc: T) -> Result<Self> {
+        self.execution_rpc = Some(
+            execution_rpc
+                .into_url()
+                .map_err(|_| eyre!("Invalid execution RPC URL"))?,
+        );
+        Ok(self)
     }
 
-    pub fn verifiable_api<T: IntoUrl>(mut self, verifiable_api: T) -> Self {
+    pub fn verifiable_api<T: IntoUrl>(mut self, verifiable_api: T) -> Result<Self> {
         self.verifiable_api = Some(
             verifiable_api
                 .into_url()
-                .expect("Invalid verifiable API URL"),
+                .map_err(|_| eyre!("Invalid verifiable API URL"))?,
         );
-        self
+        Ok(self)
     }
 
     pub fn checkpoint(mut self, checkpoint: B256) -> Self {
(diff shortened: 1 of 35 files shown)


ethereum/src/config/checkpoints.rs
```diff
@@ -151,9 +152,9 @@ impl CheckpointFallback {
         Self::fetch_latest_checkpoint_from_services(&services[..]).await
     }
 
-    async fn query_service(endpoint: &str) -> Option<RawSlotResponse> {
+    async fn query_service(endpoint: &Url) -> Option<RawSlotResponse> {
         let constructed_url = Self::construct_url(endpoint);
-        let res = get(&constructed_url).await.ok()?;
+        let res = get(constructed_url.as_str()).await.ok()?;
```
ncitron, line 157: We should make the `get` function take a Url.

ethereum/src/builder.rs
```diff
@@ -110,8 +110,8 @@ impl<DB: Database> EthereumClientBuilder<DB> {
         self
     }
 
-    pub fn fallback(mut self, fallback: &str) -> Self {
-        self.fallback = Some(fallback.to_string());
+    pub fn fallback(mut self, fallback: Url) -> Self {
```
ncitron, line 113: We should use the Into<Url> pattern that we do in some of the other builders. All builders should take Into<Url> rather than just a Url.

opstack/src/consensus.rs
```diff
@@ -179,12 +179,15 @@ fn verify_unsafe_signer(config: Config, signer: Arc<Mutex<Address>>) {
...
             }
 
+            let consensus_rpc_str = eth_config
+                .consensus_rpc
+                .as_ref()
+                .ok_or_else(|| eyre!("missing consensus rpc"))?
+                .to_string();
+
             let mut eth_consensus =
                 EthConsensusClient::<MainnetConsensusSpec, HttpRpc, ConfigDB>::new(
```
ncitron, line 189: Let's just make the consensus client take a Url too

verifiable-api/client/src/http.rs
```diff
@@ -86,8 +88,11 @@ impl<N: NetworkSpec> VerifiableApi<N> for HttpVerifiableApi<N> {
         block_id: Option<BlockId>,
         include_code: bool,
     ) -> Result<AccountResponse> {
-        let url = format!("{}/eth/v1/proof/account/{}", self.base_url, address);
-        let mut request = self.client.get(&url);
+        let url = self
+            .base_url
+            .join(&format!("eth/v1/proof/account/{}", address))
+            .unwrap();
+        let mut request = self.client.get(url.as_str());
```
ncitron, line 95: Is it not possible to pass in a url here?

PR #632, opened 2025-06-14
fix: handle RPC errors gracefully in EVM execution

### LINKED ISSUE #627: panic when making an rpc request
: thread 'tokio-runtime-worker' panicked at /build/source/core/src/execution/evm.rs:120:47:
: called `Result::unwrap()` on an `Err` value: server returned an error response: error code -32602: distance to target block exceeds maximum proof window
: Location:
:     /build/source/core/src/execution/providers/rpc.rs:213:21

kevinacker:
## Description
This PR fixes a panic that occurs when making RPC requests by properly handling error cases in the EVM execution flow. Previously, the code would unwrap RPC responses directly, causing panics when errors occurred (like exceeding the maximum proof window).

### Changes
- Modified `get_context` in `ethereum/src/evm.rs` to return `Result<Context, EvmError>`
- Replaced unwrap calls with proper error handling and propagation
- Added error context mapping for RPC errors to `EvmError::Generic`

### Testing
- Verified with `cargo check` and `cargo test`
- Error cases now return proper error types instead of panicking

### Related Issues
Fixes #627

### DIFF
--- a/core/src/execution/providers/rpc.rs
+++ b/core/src/execution/providers/rpc.rs
@@ -263,10 +263,11 @@ impl<N: NetworkSpec, B: BlockProvider<N>, H: HistoricalBlockProvider<N>> BlockPr
         // 2. Try historical provider if available and not requesting latest
         if let Some(historical) = &self.historical_provider {
             if block_id != BlockNumberOrTag::Latest.into() {
-                if let Some(block) = historical
+                let historical_result = historical
                     .get_historical_block(block_id, full_tx, self)
-                    .await?
-                {
+                    .await;
+
+                if let Ok(Some(block)) = historical_result {
                     // Note: Do NOT cache historical blocks to avoid interfering with consistency detection
                     return Ok(Some(block));
                 }
--- a/tests/rpc_equivalence.rs
+++ b/tests/rpc_equivalence.rs
@@ -984,26 +984,57 @@ async fn test_get_historical_block(helios: &RootProvider, expected: &RootProvide
     Ok(())
 }
 
+async fn test_get_too_old_block(helios: &RootProvider, expected: &RootProvider) -> Result<()> {
+    // Fetch the latest block number from the trusted provider
+    let latest_block = expected.get_block_number().await?;
+    let latest_block_num = latest_block;
+
+    // Calculate a block number that is guaranteed to be outside the EIP-2935 ring buffer
+    let old_block_num = latest_block_num - 20000;
+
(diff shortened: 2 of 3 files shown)


kevinacker:
hey @ncitron could you please take a look at this, if everything's alright

core/src/execution/providers/rpc.rs
```diff
@@ -263,10 +263,11 @@ impl<N: NetworkSpec, B: BlockProvider<N>, H: HistoricalBlockProvider<N>> BlockPr
         // 2. Try historical provider if available and not requesting latest
         if let Some(historical) = &self.historical_provider {
             if block_id != BlockNumberOrTag::Latest.into() {
-                if let Some(block) = historical
+                let historical_result = historical
                     .get_historical_block(block_id, full_tx, self)
-                    .await?
-                {
+                    .await;
```
ncitron, line 268: Let's leave it so that if the block fetch errors we return an error. Returning `None` has a special meaning of "we think there is no block here." and error should mean "we can't tell what the block is."

tests/rpc_equivalence.rs
```diff
@@ -984,26 +984,57 @@ async fn test_get_historical_block(helios: &RootProvider, expected: &RootProvide
     Ok(())
 }
 
+async fn test_get_too_old_block(helios: &RootProvider, expected: &RootProvider) -> Result<()> {
+    // Fetch the latest block number from the trusted provider
+    let latest_block = expected.get_block_number().await?;
+    let latest_block_num = latest_block;
+
+    // Calculate a block number that is guaranteed to be outside the EIP-2935 ring buffer
+    let old_block_num = latest_block_num - 20000;
```
ncitron, line 993: let's use `saturation_sub` here to prevent underflows if we run these tests on a new network (like a testnet)

tests/rpc_equivalence.rs
```diff
@@ -984,26 +984,57 @@ async fn test_get_historical_block(helios: &RootProvider, expected: &RootProvide
...
+    );
+
+    Ok(())
+}
+
 async fn test_get_historical_balance(helios: &RootProvider, expected: &RootProvider) -> Result<()> {
-    let latest = helios.get_block_number().await?;
-    let historical_block_num = latest.saturating_sub(100);
+    let latest_block = helios.get_block_number().await?;
+    let target_block_num = latest_block - 100;
```
ncitron, line 1021: go black to `saturating_sub`

PR #633, opened 2025-06-16
Remove deprecated CDN links from README

Daulox92:
Remove non-functioning unpkg.com CDN links from the documentation. These links are no longer accessible as the package is not available through CDN.
Changes:
Removed https://unpkg.com/@a16z/helios/dist/lib.umd.js from UMD example
Removed https://unpkg.com/@a16z/helios/dist/lib.mjs from ESM example
Updated import statements to use empty source paths


ncitron:
these links will be created soon

PR #650, opened 2025-07-07
expose eyre

zemse:
Currently dependents have to add a `eyre` dependency for [handling](https://github.com/zemse/gm/blob/abeaccaf058d83358faba08dacee140871b5bc51/src/error.rs#L50) the errors.

Exposing the error library would be very helpful.

### DIFF
--- a/ethereum/src/lib.rs
+++ b/ethereum/src/lib.rs
@@ -13,3 +13,4 @@ mod constants;
 
 pub use builder::EthereumClientBuilder;
 pub type EthereumClient = HeliosClient<Ethereum>;
+pub use eyre;


zemse:
> Otherwise we'd end up wanting to re-export a whole host of things such as alloy.

I agree that exporting the entire eyre module is too much. 

But I think there is value in exporting at least the error type because public functions in the helios repo return a `Result<SomeType, eyre::Report>`. This was the original motivation while proposing this change.

```rs
// in helios
pub use eyre::Report as Error;
```

> and don't think its a particular burden to import eyre yourself

Usually it's not a burden but dependent have to worry about getting the version right.

```
// in dependent
eyre = "*"
```

Above code in dependent work in most cases, until there's a strange case where are two dependencies that use different versions of eyre but don't report the error part used in the `Result`.

PR #651, opened 2025-07-08
WorkerProvider Implementation and Example

AustinFoss:
A couple months ago I started experimenting with using Helios inside a worker thread. Given the resources required for Helios' proving system, being able to offload its processes to a different thread can have several benefits. Mainly UX speed and Time To Interactive metrics, as well as removing the possibility of blocking the main thread.

Using libraries like viem and ethers inside the worker is trivial, especially after v0.9.0 release making Helios integration through wasm in a JS environemnt ( you guys made that so much more smooth), but it means having to rewrite the logic of the worker for each use case.

Using Helios inside a worker but making ETH_RPC calls from the main thread I find gives a lot more flexibility in terms of not needing to mess with a file that needs to be pre-bundled and either deliverable through a CDN, or present in the project's public directory.

What I've attempted to do is implement an agnostic wrapper for an instance of a Worker making the Worker instance EIP-1193 compatible. The wrapper handles communication of the EIP-1193 req/res to and from the worker that's compliant with EIP-1193.

The worker is structured in a way that lets it handle multiple networks.  For instance, the way I modified the example is to have Optimism-mainnet continue in the main thread, Ethereum on it's own in one worker, while both Base and Linea in a single worker together.

There's two things that are missing I'd like to add before this gets accepted (if it does), I just want to test the waters with how this proposal is received before going further:
- more documentation in JSDOC format and a further elaboration in the README instructions
- IndexedDB caching for the checkpoint because localstorage isn't accessible from a worker.

## Summary

The only changes to the main library are the additions of `WorkerProvider`, `createWorkerProvider()`, & `NetworksCRUD` to `helios-ts/lib.ts`. 

Followed by some change to package.json and a new vite config file to handle the worker compilation for the example, plus further file modifications to the example directory to demonstrate how this is implemented.

The worker.ts file is meant for demonstration purposes. As was mentioned in telegram when I first brought this up, distributing a pre-bundled file through npm is a little tricky and might best be left up to each developer to implement their own worker script, but this example could be a good baseline. If this is accepted I'm sure there is a way to bundle the worker.ts file and have it delivered by CDN.

### DIFF
--- a/helios-ts/example/src/main.ts
+++ b/helios-ts/example/src/main.ts
@@ -145,7 +145,7 @@ async function main() {
   networks.forEach((n) => {
     setInterval(async () => {
       try {
-        if (!n.viemClient) return;        
+        if (!n.viemClient) return;
         
         const latestNumber = await n.viemClient.getBlockNumber();
         if (latestNumber !== n.lastSeen) {


helios-ts/example/src/worker/worker.ts
```diff
@@ -0,0 +1,115 @@
...
+            console.error('Unhandled message/<method>:', e.data.method);
+    }
+};
+
+const handleNetworks = async (options: {method: string, params: {name?: string, cfg: Config, kind: NetworkKind}}): Promise<any> => {
+    const { method, params: { name = '', cfg, kind } } = options; // Set default value for name
+    // const method = options.method
+    // const params = options.params
+    
+    return new Promise(async (resolve, reject) => {
```
karen-sarkisyan, line 35: Here and in other instances of the PR you're using an `async` function -- that always returns a promise -- along with explicitly returning a `Promise`. While it's not a mistake, a cleaner approach would be to just stick to `async`, returning value and `throw`ing in try/catch block

ncitron:
Yeah I agree a lot with @karen-sarkisyan's point that we probably want to give users a working provider to start with. They can of course roll their own worker setup if our default doesn't suit them, but it's not clear to me why a standard one would't work in 99% of cases.

PR #653, opened 2025-07-09
docs: broken link

avorylli:
### Not significant change 

deleted 404 link and made it as it is at the line 30


ncitron:
We will be revamping the README shortly so these changes won't be necessary. 

We also generally do not accept trivial PRs to discourage airdrop farming.

PR #655, opened 2025-07-11
refactor(bls): use idiomatic ok_or_else for point conversion

Galoretka:
Refactored the point methods in both PublicKey and Signature to use .ok_or_else(|| eyre!("invalid point")) instead of manual is_some() and unwrap() checks when converting from compressed bytes. This makes the code more idiomatic, concise, and safer by avoiding unnecessary unwraps and improving error handling.


ncitron:
Does not compile and does not include meaningful changes.

PR #663, opened 2025-07-23
fix: improve error handling in HTTP client initialization

reallesee:
Replace unwrap() with proper error handling when creating HTTP client in poller.rs. This change prevents potential application crashes if client creation fails (e.g., due to network issues), making the system more robust


ncitron:
We are not accepting these kinds of PRs at this time to discourage airdrop farming.

PR #672, opened 2025-08-05
patch unwrap

zemse:
I encountered a panic and it came from this one. It would be great if it can be bubbled up by Err, so that I can spawn a fresh helios instance again once the network is back up.


ncitron (commented):
Looks good, but we should implement this for both opstack and ethereum.

PR #689, opened 2025-08-21
fix potential panic on empty storage_proof array

prestoalvarez:
Replace unsafe array indexing `[0]` with safe `first()` access in OP Stack consensus verification to prevent process crashes when receiving empty `storage_proof` from RPC responses.


prestoalvarez:
@ncitron

PR #731, opened 2025-11-03
fix: disable tx validation for estimateGas and createAccessList

karen-sarkisyan:
Both `eth_estimateGas` and `eth_createAccessList` should not validate gas limit, base fee, eip3607 and nonce, because it is a local execution for estimation and simulation purposes, much like `eth_call`.

It is consistent with Geth and Reth behavior.
https://github.com/paradigmxyz/reth/blob/0bca7b150db529ac41616a1d542aa0239fb57b99/crates/rpc/rpc-eth-api/src/helpers/estimate.rs#L50-L81

https://github.com/paradigmxyz/reth/blob/1f2f1d432f9420916cabe5e6d0623b8b92a18ea7/crates/rpc/rpc-eth-api/src/helpers/call.rs#L409-L426

I propose we don't remove the flag entirely though, in case it will be needed in the future to handle other methods.

### DIFF
--- a/core/src/client/node.rs
+++ b/core/src/client/node.rs
@@ -200,7 +200,7 @@ impl<N: NetworkSpec, C: Consensus<N::BlockResponse>, E: ExecutionProvider<N>> He
 
         let (result, ..) = N::transact(
             tx,
-            true,
+            false,
             self.execution.clone(),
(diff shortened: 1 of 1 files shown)


ncitron (approved):
LGTM!

PR #733, opened 2025-11-03
fix: raw transaction decode

karen-sarkisyan:
Usually, there's a `0x` prefix in transaction data but here we don't strip it:
https://github.com/a16z/helios/blob/daffbbbf4367a1fb1baff717ebc3452b7d950909/helios-ts/src/ethereum.rs#L355-L359 

`hex` lib used here is 0.4.3 which [does not strip](https://docs.rs/hex/0.4.3/hex/fn.decode.html) `0x` prefix itself, which causes a bug for us.

The `hex` re-exported by alloy, [does](https://docs.rs/const-hex/1.14.1/const_hex/fn.decode.html).

In this PR I removed the unnecessary dated `hex` dependency and use alloy's re-export instead, fixing raw transaction submission.

### DIFF
--- a/helios-ts/Cargo.toml
+++ b/helios-ts/Cargo.toml
@@ -19,7 +19,6 @@ eyre.workspace = true
 alloy.workspace = true
 op-alloy-rpc-types.workspace = true
 
-hex = "0.4.3"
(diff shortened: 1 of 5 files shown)


ncitron (approved):
LGTM!

PR #753, opened 2025-12-30
chore: resolve benchmark compilation errors

mattsse:
Fixes compilation errors in benchmark files.

Updated benchmarks to use `BlockId` instead of removed `BlockTag` type, added required `with_file_db()` calls to builder chains, and replaced deprecated `into_path()` with `keep()`.


ncitron (approved):
LGTM!

PR #757, opened 2026-01-09
perf: skip unchanged finalized block fetch and send

karen-sarkisyan:
This PR notably improves performance of mainnet advance tracking algorithm:
- Skips finalized block fetch when it hasn't changed according to finality update
- Finalized+optimistic block fetching is done in parallel instead of sequentially

Performance assessment:
While the result is highly sensitive to the number of transactions in the block and network speed, ballpark figures of perf improvement of mainnet advance algorithm would be something like this:
1. ≈2x-3x faster latest (optimistic) block update due to skipped finalized block fetching and recovery
2. ≈30% faster finalized block update due to parallelized fetching
3. ≈45% less main thread blocking time while tracking tip of chain which can amount to ≈15 seconds of total main thread blocking per epoch in WASM (assuming 80ms blocking finality update verifications after https://github.com/a16z/helios/pull/751 is applied and 500ms single `payload_to_block` call blocking time) due to skipped finalized block recoveries when unchanged

### DIFF
--- a/ethereum/src/consensus.rs
+++ b/ethereum/src/consensus.rs
@@ -475,15 +475,43 @@ impl<S: ConsensusSpec, R: ConsensusRpc<S>> Inner<S, R> {
 
     pub async fn send_blocks(&self) -> Result<()> {
         let slot = self.store.optimistic_header.beacon().slot;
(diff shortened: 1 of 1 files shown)


ncitron (approved):
LGTM!

PR #761, opened 2026-01-12
perf: skip unnecessary eth_getCode requests

karen-sarkisyan:
This PR does 2 optimizations for EVM calls regarding data fetching:

- Disable eth_getCode request on every storage request. Currently, every fetch of storage slot causes a `eth_getCode` call for the account
- Skip code fetching completely when `code_hash` is nullish in the proof, meaning there's no code to fetch

Also updated `get_basic` to mark account data as missing when code is `None`. Not sure if the case of it being `None` is probable, but we `unwrap` on it in a few places, so it's a bit of hardening for unexpected things (like if REVM only touches a storage slot but somehow not the account itself previously).


ncitron (approved):
LGTM!

PR #770, opened 2026-01-21
fix: missing execution_forks in WASM client config

zeroprooff:
The WASM client was not copying execution_forks from the base network config, falling back to ForkSchedule::default() which sets all fork timestamps to u64::MAX. This caused the EVM to run in FRONTIER mode instead of the correct fork (Cancun/Prague), breaking eth_call, eth_estimateGas and eth_createAccessList results.


ncitron (approved):
LGTM!

PR #771, opened 2026-01-24
feat: execution data caching for EVM

karen-sarkisyan:
Ths PR implements basic caching. It is essentially EVM caching but also the first step to global caching of execution RPC data.

#### General info
- It's an **in-memory** cache.
- Eviction strategy is handled by `schnellru` crate.
- Additional changes: type-safe block hash parameter for `ProofDB` and removed unconditional call of `prefetch_state` without first fetching cache.

From architecture point of view, the main challenge is combining **synchronous reads by EVM and asynchronous fetching when cache is missed**. Which is why `ProofDB` has direct access to the inner `cache` and can bypass `CachingProvider` that wraps Execution RPC.

#### Rules
Caching is built on 3 core rules:
1. Account proofs must be fetched fresh for a given block, proof can only be reused for the same block hash
2. Storage slots are cached by storage root hash, can be reused between blocks
3. Contract code is cached by code hash and shared across accounts, can be reused between blocks

## Performance profiling
- I did some profiling using requests from [Kohaku fork of Ambire](https://github.com/ethereum/kohaku-extension) as examples
- Timings are naturally sensitive to network latency. At the time of testing my ankr RPC provider was not very fast
- Profiling was done in JSON-RPC server mode

A getBalance request 

| Req# | Time     | EVM loop iterations | Network requests |
| ---- | -------- | ------------------- | ---------------- |
| 1    | 4.3s     | 12                  | 15               |
| 2    | **3.6s** | 12                  | **12**           |

This call is touching only contracts with lots of changes every block. However, caching still helped trim 0.7s and network costs thanks to skipping `getCode` requests.

A UniswapV3 position request

| Req# | Time     | EVM loop iterations | Network requests |
| ---- | -------- | ------------------- | ---------------- |
| 1    | 10.2s    | 30                  | 36               |
| 2    | **4.9s** | 18                  | **18**           |

Here the win is much clearer as it makes **2x fewer** network requests and thus completes the request **2x faster**.

## Considerations
When interacting with highly active contracts (which is probably common), the win may be quite small. However, I still believe it's valuable because:
- Same requests at the same block are **near-instant** with 0 network requests. (For example, second Uniswap request targeting the same same block takes about 10ms to complete)
- Contract code is not re-fetched unless its hash changed
- It's a ground work for future improvements to better handling of multiple concurrent requests and potentially high load.

### DIFF
--- a/core/src/execution/cache/state.rs
+++ b/core/src/execution/cache/state.rs
@@ -105,10 +105,38 @@ impl Cache {
         code.get(&code_hash).cloned()
     }
 
-    /// Insert code into the cache by code hash. Hash MUST come from a verified account proof.
-    pub fn insert_code(&self, code_hash: B256, code: Bytes) {
-        let mut code_cache = self.code.write().unwrap_or_else(|e| e.into_inner());
-        code_cache.insert(code_hash, code);
+    /// Try to recover code for an address from any cached account entry.
+    ///
+    /// This is intentionally "optimistic": we only use the code if a freshly
+    /// fetched proof confirms the same code hash.
+    pub fn get_code_optimistically(&self, address: Address) -> Option<(B256, Bytes)> {
+        let code_hash = {
+            let accounts = self.accounts.read().unwrap_or_else(|e| e.into_inner());
+            accounts.iter().find_map(|((cached_address, _), account)| {
+                (*cached_address == address).then_some(account.code_hash)
+            })?
+        };
+
+        let code = self.get_code(code_hash)?;
+        Some((code_hash, code))
+    }
+
+    /// Insert a verified [`Account`] into the cache, distributing its data to
+    /// the account-proof, storage-proof, and code sub-caches.
+    pub(crate) fn insert_account(&self, address: Address, account: &Account, block_hash: B256) {
+        self.insert(
+            EIP1186AccountProofResponse {
+                address,
+                balance: account.account.balance,
+                code_hash: account.account.code_hash,
+                nonce: account.account.nonce,
+                storage_hash: account.account.storage_root,
+                account_proof: account.account_proof.clone(),
+                storage_proof: account.storage_proof.clone(),
+            },
+            account.code.clone(),
+            block_hash,
+        );
     }
 
     /// Get an account proof response with requested storage slots.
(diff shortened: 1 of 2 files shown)


common/src/execution_provider.rs
```diff
@@ -23,6 +27,10 @@ pub trait ExecutionProvider<N: NetworkSpec>:
     + Sync
     + 'static
 {
+    fn state_cache(&self) -> Arc<dyn StateCache> {
+        static NOOP: OnceLock<Arc<NoopStateCache>> = OnceLock::new();
```
ncitron: Why is this static?

ncitron (approved):
My main question is why the presence of a cache leaks into the implementation of ProofDB? Shouldn't it be possible to encapsulate the cache entirely in the `AccountProvider`?


This would mean that as long as the account provider returns cached data when appropriate the `ProofDB` does not need to worry about other details?

core/src/execution/cache/state.rs
```diff
@@ -0,0 +1,361 @@
...
+// Each entry is a HashMap of slots per storage root.
+const STORAGE_CACHE_SIZE: u32 = 128;
+
+// Code: Static and can be shared. Most valuable cache.
+const CODE_CACHE_SIZE: u32 = 300;
+
+pub struct Cache {
+    /// Storage proofs: content-addressed by storage_hash
+    /// storage_hash -> slot -> full storage proof
+    storage: RwLock<LruMap<B256, HashMap<B256, EIP1186StorageProof>>>,
```
ncitron: Also are we concerned that the size of this can blow up if we have a small number of contracts (so it fits in the `LruMap`) but a giant number of individual storage proofs in each one?

karen-sarkisyan:
> My main question is why the presence of a cache leaks into the implementation of ProofDB? Shouldn't it be possible to encapsulate the cache entirely in the `AccountProvider`?
> 
> This would mean that as long as the account provider returns cached data when appropriate the `ProofDB` does not need to worry about other details?

That is actually a good call. I traded cleanness for performance (to expose direct synchronous access to cache for `ProofDB`) but the win wasn't probably worth it. Made a deep rework. All the logic is now in `AccountProvider` and I removed `CachingProvider` entirely.

common/src/execution_provider.rs
karen-sarkisyan: Removed this change entirely

common/src/state_cache.rs
```diff
@@ -0,0 +1,36 @@
...
+/// implementation without requiring async.
+pub trait StateCache: Send + Sync + 'static {
+    fn get_storage(&self, address: Address, slot: B256, block_hash: B256) -> Option<U256>;
+
+    fn get_account(
+        &self,
+        address: Address,
+        slots: &[B256],
+        block_hash: B256,
+    ) -> Option<(Account, Vec<B256>)>;
```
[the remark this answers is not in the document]
karen-sarkisyan: Great observation, added a comment

PR #777, opened 2026-02-05
chore (TS): update wasmtimer dependency

karen-sarkisyan:
- updated `wasmtimer` version and moved to root
- removed unused `zduny-wasm-timer` and `gloo-timers`

This also seems to have fixed a weird issue of multiple 5-second `setTimeout`s constantly being called without doing notable work


ncitron (approved):
LGTM!

PR #780, opened 2026-02-09
fix: pin `get_code_at` to proof block in `RpcExecutionProvider`

Galoretka:
## Why

`get_code_at` in [RpcExecutionProvider::get_account]https://github.com/a16z/helios/blob/4a32ac1a9fbcf46386a497e4e0a7232ad1388762/verifiable-api/server/src/service.rs#L58-L111 was called without `.block_id()`, defaulting to `latest`. Meanwhile `get_proof` right above
it was pinned to a specific block hash. If a new block arrived between the two RPC calls, the returned code could belong to a different state than the proof, causing a spurious `CodeHashMismatch` error.

## What

Added `.block_id(block.header().hash().into())` to the `get_code_at` call so it queries the same block as `get_proof`. This matches how every other callsite in the codebase already uses `get_code_at`.


ncitron (approved):
LGTM!

PR #782, opened 2026-02-13
perf(bls): remove redundant G2 subgroup check in signature verification

sashass1315:
The [Signature::verify()](https://github.com/a16z/helios/blob/4a32ac1a9fbcf46386a497e4e0a7232ad1388762/ethereum/consensus-core/src/types/bls.rs#L39-L50 method was performing an explicit subgroup check via
[subgroup_check_g2()]https://github.com/a16z/helios/blob/4a32ac1a9fbcf46386a497e4e0a7232ad1388762/ethereum/consensus-core/src/types/bls.rs#L84-L90 after calling `G2Affine::from_compressed()`. However, `from_compressed()` already performs a subgroup check internally via `is_torsion_free()` - this is a documented property of the bls12_381 crate.
The explicit check was performing an expensive scalar multiplication (`point * r` where r is the 255-bit curve order) that duplicated work already done during decompression.
 

Remove the redundant [subgroup_check_g2()] function and its call from [Signature::verify()]. Also remove the now-unused [hex_to_scalar()] helper and the `Scalar` import. This eliminates one full G2 scalar multiplicationper signature verification (~30-50% speedup for the verification path).

### DIFF
--- a/ethereum/consensus-core/src/types/bls.rs
+++ b/ethereum/consensus-core/src/types/bls.rs
@@ -1,6 +1,6 @@
 use bls12_381::{
     hash_to_curve::{ExpandMsgXmd, HashToCurve},
-    multi_miller_loop, G1Affine, G1Projective, G2Affine, G2Prepared, G2Projective, Gt, Scalar,
+    multi_miller_loop, G1Affine, G1Projective, G2Affine, G2Prepared, G2Projective, Gt,
 };
 use eyre::{eyre, Result};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


sashass1315:
bench rsults:
before:
image
after: 
image

sashass1315:
> For context, a similar optimization has been postponed, see here [#751 (comment)](https://github.com/a16z/helios/pull/751#issuecomment-3732802239)

Ohh sorry, didn't really thought about looking at PRs. So I close this PR then? if it's yours idea

PR #788, opened 2026-03-24
`save_new_checkpoints` can busy-wait if ConsensusClient is incorrectly dropped

### LINKED ISSUE #787: `save_new_checkpoints` can busy-wait if ConsensusClient is incorrectly dropped
I encountered an issue with helios-ts where helios would busy-wait in `consensus::save_new_checkpoints` indefinitely.

I believe the error is triggered when `ConsensusClient` gets dropped (possibly garbage collected?), which drops `shutdown_send`.  Once all its senders are dropped, the `shutown_rx` seems to return an `Err()` for `borrow` or `borrow_and_update`.  Since `borrow` only checks the most recently seen value, not any error value, it gets stuck.

```rust
_ = shutdown_rx.changed() => {
    if *shutdown_rx.borrow() { // if (shutdown_rx.is_err() == true), we have no way of breaking
        info!(target: "helios::consensus", "shutting down consensus client");
        break;
    }
}
```

Fix is to also break if `shutdown_rx.is_err()`. This will only be true if all shutdown_tx instances are dropped, which should never happen during regular operations.

I'll attach a quick PR with the fix.

Robert-MacWha:
Fixes https://github.com/a16z/helios/issues/787?reload=1

### DIFF
--- a/ethereum/src/consensus.rs
+++ b/ethereum/src/consensus.rs
@@ -179,8 +179,8 @@ impl<S: ConsensusSpec, R: ConsensusRpc<S>, DB: Database> ConsensusClient<S, R, D
 
             loop {
                 tokio::select! {
-                    _ = shutdown_rx.changed() => {
-                        if *shutdown_rx.borrow() {
+                    result = shutdown_rx.changed() => {
+                        if result.is_err() || *shutdown_rx.borrow_and_update() {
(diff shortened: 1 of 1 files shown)


ncitron (approved):
LGTM!

PR #794, opened 2026-05-08
feat: forward Authorization header to upstream RPC

markdavison:
## Summary

Adds optional forwarding of the inbound `Authorization` header to upstream RPC calls. Default off; opt in by setting `HELIOS_FORWARD_AUTH=1`.

This is useful when Helios is deployed behind a reverse proxy or gateway that authenticates each request and the upstream execution-RPC backend expects the same credentials. Without this, operators have to either (a) configure a static credential on the `--execution-rpc` URL, which only supports basic auth and can't propagate per-request identity, or (b) terminate auth in front of Helios with a custom shim that stitches credentials back onto upstream calls.

## Implementation

- **Inbound** (`core/src/auth_forwarding::AuthCaptureLayer`): tower middleware on the jsonrpsee server captures the `Authorization` header into a `tokio::task_local!` for the request scope.
- **Outbound** (`core/src/auth_forwarding::AuthForwardLayer`): an alloy `RpcClient` layer reads the task-local and writes an `Authorization` header into the request-meta `HeaderMap` extension. The existing `Http<reqwest::Client>` transport already reads `req.headers()` and forwards on the upstream HTTP request, so no transport changes were needed.
- **Both** `RequestPacket::Single` and `RequestPacket::Batch` are handled.

## HTTP-only

jsonrpsee dispatches WebSocket frames on freshly spawned tasks that don't inherit the request's task-local. Rather than silently dropping auth on every WS frame, `jsonrpc::start` calls `.http_only()` whenever the feature is enabled. The behavior is documented on the module.

## Testing

Six unit tests in `core/src/auth_forwarding::tests`:

- `capture_sets_task_local_when_header_present`
- `capture_records_none_when_header_absent`
- `forward_injects_header_when_task_local_set`
- `forward_injects_header_into_each_request_in_batch`
- `forward_is_noop_when_task_local_unset`
- `is_enabled_reads_env_var` (`#[serial_test::serial]`)

`cargo clippy -p helios-core --all-targets -- -D warnings` is clean.

## Configuration

Default off — no behavior change unless `HELIOS_FORWARD_AUTH=1` is set in the environment. Happy to convert this to a CLI flag (`--forward-auth-header`) if you'd prefer that style.


markdavison:
Closing for now — the `.http_only()` workaround drops `eth_subscribe(newHeads)` support. Will revisit once that's addressed.

PR #795, opened 2026-05-16
small typo fix

kevaundray:


karen-sarkisyan:
don't think this fix is necessary, "unmanipulable" is an actual word: https://en.wiktionary.org/wiki/unmanipulable

kevaundray:
fair, will close :)

Issue #302, opened 2023-11-20
feat: support tracking consensus updates in helios-ts 


rdvorkin:
Support in some way tracking and getting notified when a new update is applied. The usecase I want to implement is setting a hook on new optimistic update, to do some action (For example, re-check a list of balances) when the update is applied.

Implementing something like https://nodejs.org/api/events.html#emitteroneventname-listener would be ideal, however even a function returning the current status (latest update applied, current block etc) would be great.

ncitron:
Would something that mimics the websocket block subscription api make sense here potentially?

rdvorkin:
@ncitron Yeah I suppose so, could you link an example of how that would look like?

rdvorkin:
@ncitron Yes this seems perfect!

ppca:
Hey I'm wondering if there's support for `eth_subscribe` in the rust crate now?

ppca:
Thanks @karen-sarkisyan !

ncitron:
@karen-sarkisyan Is this for the eth_subscribe newHeads or the filter one? We already support tracking new heads for a filter, but have no subscribe support so if thats it feel free.

ppca:
@ncitron What does `tracking new heads for a filter` mean?
What I want to do is basically 
```
web3_ws.eth_subscribe().subscribe_logs(filter)
```
Do you already have support to enable this?

PR #798, opened 2026-05-23
feat: optimistic block notifications

pauldelucia:
## Summary

Adds a `watch`-channel notification path for the latest optimistic-header-derived block, mirroring the shape of #738's checkpoint notifications. Builds on the consensus client's already-tracked `optimistic_header` state — no new sync logic, just exposes data that the light client already maintains.

```rust
// new on the Consensus trait
fn optimistic_block_recv(&self) -> Option<watch::Receiver<Option<B>>>;

// new on HeliosApi
async fn current_optimistic_block(&self) -> Result<Option<N::BlockResponse>>;
fn new_optimistic_blocks_recv(&self) -> Result<watch::Receiver<Option<N::BlockResponse>>>;
```

## Motivation

#302 originally asked for "tracking and getting notified when a new update is applied" with a concrete use case of "setting a hook on new optimistic update, to do some action (e.g. re-check a list of balances)". #738 addressed the checkpoint half of that — the finalized side. This addresses the optimistic side.

`block_recv` already streams the same blocks via mpsc, but the watch variant lets multiple subscribers each see the latest value without coordinating a single consumer. Concrete external use cases:

- A data indexer that needs to react to chain-tip advances under sync-committee attestation without waiting for full finality (~12.8min).
- A dashboard / SSE service that fans out updates to many connected clients — each subscriber holds its own `watch::Receiver` clone of the same channel.
- A monitoring tool tracking "what does Helios think is the head right now?" for health checks.

For comparison: `block_recv` (mpsc) is a stream meant for a single consumer; the new watch channel is meant for many lightweight observers that only care about the latest value.

## Changes

- `core/src/consensus.rs` — add `optimistic_block_recv` to the `Consensus` trait
- `core/src/client/api.rs` — add `current_optimistic_block` + `new_optimistic_blocks_recv` to `HeliosApi`
- `core/src/client/node.rs` — implement the new `HeliosApi` methods (delegate to `Consensus::optimistic_block_recv`)
- `ethereum/src/consensus.rs` — wire `optimistic_block_send` into `ConsensusClient` + `Inner`, publish in both branches of `send_blocks` (matching how `block_send` was already invoked)
- `opstack/src/consensus.rs`, `linea/src/consensus.rs` — implement the trait method returning `None`, matching how they already handle `checkpoint_recv`

44 lines added across 6 files. `cargo test --workspace --lib` clean (23 tests passing), `cargo clippy --workspace --lib` clean, `cargo fmt --check` clean.

## What's deliberately NOT in this PR

- **JSON-RPC method + subscription.** Would mirror `helios_getCurrentCheckpoint` / `helios_subscribeNewCheckpoints` from #738, but doing so cleanly requires making the `HeliosRpc` trait generic over `N::BlockResponse` (the current shape returns concrete `Option<B256>` for checkpoints, which works because B256 is concrete; blocks are network-specific). That's a bigger trait-surface change that the maintainers may prefer to design themselves. Happy to follow up in a separate PR if you want this — the Rust-API additions in this PR work standalone for crate consumers.
- **`helios-ts` bindings.** Depend on the JSON-RPC layer, so deferred along with that.

## Note on tests

There's no test directly observing the new channel — there's no analogous test for `finalized_block_send` either, since exercising `send_blocks` end-to-end needs richer mock-RPC fixture work than the existing tests do. The channel wiring is mechanical (mirrors `finalized_block_send` exactly). Happy to add a test if reviewers prefer — pointer to the right test pattern would help.

### DIFF
--- a/core/src/client/api.rs
+++ b/core/src/client/api.rs
@@ -90,4 +90,9 @@ pub trait HeliosApi<N: NetworkSpec>: Send + Sync + 'static {
     // checkpoint
     async fn current_checkpoint(&self) -> Result<Option<B256>>;
     fn new_checkpoints_recv(&self) -> Result<tokio::sync::watch::Receiver<Option<B256>>>;
+    // optimistic block (head attested by current sync committee)
+    async fn current_optimistic_block(&self) -> Result<Option<N::BlockResponse>>;
+    fn new_optimistic_blocks_recv(
+        &self,
+    ) -> Result<tokio::sync::watch::Receiver<Option<N::BlockResponse>>>;
 }
--- a/core/src/client/node.rs
+++ b/core/src/client/node.rs
@@ -522,4 +522,19 @@ impl<N: NetworkSpec, C: Consensus<N::BlockResponse>, E: ExecutionProvider<N>> He
             .checkpoint_recv()
             .ok_or_else(|| eyre!("Checkpoints not supported"))
     }
+
+    async fn current_optimistic_block(&self) -> Result<Option<N::BlockResponse>> {
+        self.consensus
+            .optimistic_block_recv()
+            .map(|recv| recv.borrow().clone())
+            .ok_or_else(|| eyre!("Optimistic block updates not supported"))
+    }
+
+    fn new_optimistic_blocks_recv(
+        &self,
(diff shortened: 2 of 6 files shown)


pauldelucia:
Hi @ncitron could you have a look at this one please?

PR #805, opened 2026-06-24
feat: glamsterdam support

ncitron:


karen-sarkisyan:
FYI: root `alloy` has been updated to `v2.1` so there's a problem compiling in WASM because of `getrandom`'s newer version `0.4.3`. I think this line should be updated in `core/Cargo.toml` to enable "wasm_js" feature for the other version:

https://github.com/a16z/helios/blob/844d01323f25432d6c28b3e3bff567705597aaab/core/Cargo.toml#L43

fakedev9999:
Hi @ncitron, we’re assessing Glamsterdam support for [sp1-helios](https://github.com/succinctlabs/sp1-helios) and comparing this branch against the current Gloas spec.

I noticed the light-client proof paths still differ from the [current spec](https://github.com/ethereum/consensus-specs/blob/fda6c52f94d0d91eb3f64d4ef5d510865a8ec73f/specs/gloas/light-client/sync-protocol.md#L108). For example, the finality branch uses 7 hashes here versus 9 in the spec, and the execution block hash branch uses 9 versus 11.

Are updates for these changes planned in this PR, or is there a newer branch we should follow? Any rough timing would help us plan our integration and assess where we could contribute.

ncitron:
Hey @fakedev9999

I've updated this branch to comply with the latest specification. We'll likely merge this sometime this week, then take some time do a security review of the changes before cutting a new release next week.

If you run into any issues with these changes please let me know!

PR #810, opened 2026-09-08
fix: return an error instead of panicking on a receipt or tx without an index

kriss39:
`verify_receipt_proof` and `verify_transaction_proof` unwrapped `transaction_index()`:

```rust
let index = receipt.transaction_index().unwrap() as usize;
```

That field is `None` for anything not included in a block, and both functions are called on responses that come straight off a verifiable-api server, before any of that server's data has been proven:

```rust
// providers/verifiable_api.rs
let Some(block_hash) = receipt_response.receipt.block_hash() else { ... };
// ...
verify_receipt_proof::<N>(&receipt_response.receipt, receipts_root, &receipt_response.receipt_proof)?;
```

Only `block_hash` is checked there. A response with `"transactionIndex": null` panics, and since release builds set `panic = "abort"` that takes the whole process down. It is the wrong outcome for a function whose job is to decide whether to trust that server in the first place.

`get_transaction_by_location` already handles the same Option properly with `ok_or(eyre!("transaction not included"))`, so this is really the two call sites that were missed.

Both now return `ExecutionError::TransactionNotIncluded` and the response gets rejected like any other invalid one.

Added two tests. I checked they fail on the current code for the right reason — both panic at `proof.rs` with `called Option::unwrap() on a None value` — and pass after the change.

`cargo fmt --all --check` is clean and the change introduces no new clippy findings. `cargo test -p helios-core` passes (16). Two tests in the workspace fail for me both with and without this change, since I have no credentials for them: `rpc_equivalence` needs `MAINNET_EXECUTION_RPC`, and `checkpoints::test_fetch_latest_checkpoints` calls out to live checkpoint services.

The same pattern exists in a few other spots I noticed while reading (`storage_proof[0]` in the opstack unsafe-signer check, the `TxEnvelope::decode().unwrap()` calls in `payload_to_block`). Happy to send those separately if you want them fixed.

### DIFF
--- a/core/src/execution/errors.rs
+++ b/core/src/execution/errors.rs
@@ -21,6 +21,8 @@ pub enum ExecutionError {
     MissingLog(B256, U256),
     #[error("too many logs to prove: {0} spanning {1} blocks current limit is: {2} blocks")]
     TooManyLogsToProve(usize, usize, usize),
+    #[error("transaction {0} is not included in a block")]
+    TransactionNotIncluded(B256),
     #[error("execution rpc is for the incorrect network")]
     IncorrectRpcNetwork(),
     #[error("block not found: {0}")]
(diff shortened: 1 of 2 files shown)


ncitron (approved):
LGTM!

